PR #88, opened 2014-07-01
@_cancan_skipper should be inherited: add failing test

jpmckinney:
If a superclass controller has:

```
load_and_authorize_resource
skip_load_resource :only => :index
```

A subclass will not skip the loading of the resource on the `index` action.

I've added a failing test to illustrate this behavior.

Potential fix would be to use `class_attribute` from Rails: http://apidock.com/rails/Class/class_attribute

Moved from https://github.com/ryanb/cancan/pull/683


coorasse:
@jpmckinney would you be up to fix this and send a PR?

jpmckinney:
@coorasse When running `bundle exec rake spec` on a clean `develop` branch, I get a lot of errors, because ActiveSupport isn't loaded?

coorasse:
check the `Development` section of the README: https://github.com/CanCanCommunity/cancancan#development

cancancan uses Appraisals

jpmckinney:
@coorasse Thanks! I think my new commit will get the failing test passing.

coorasse:
The CI is still failing. Fix it and then we can schedule this PR for the next version

jpmckinney:
@coorasse They look like Rubocop failures for which I'm not responsible (I didn't touch those files).

jpmckinney:
Ahhh, nevermind, I see what I did.

PR #116, opened 2014-08-28
Fix #115 - specify custom parent authorization action.

### LINKED ISSUE #115: Provide a custom authorization action on parent resources
Currently, when authorizing nested resources, the parent resource is always authorized with the :show action. This makes a great deal of sense and is a good default but there are occasions when are more restrictive option is needed for more complex permissions.

I'd like to be able to specify the action that cancan uses to authorize the parent object as well 

``` ruby
class Controller
  load_and_authorize_resource :library, parent: true, parent_action: :sort
end
```

phallguy:


phallguy:
@bryanrite Do you need anything else from me to get this merged in?

Fryie:
+1 :)

Issue #105, opened 2014-07-30
Feature request: add if: and unless: to hash options


fabn:
This is a feature request for my use case, I don't know if it would be useful to all. I'm trying to cache generated `Ability` class taking inspiration from [this gist](https://gist.github.com/jackquack/5718023).

However in that gist the user skips block conditions and recreate them during `marshal_load` but that would interfere with condition order and it's not scalable.

In my case I've written all conditions in hash form and the only conditions I still need to make cacheable have this form

``` ruby
can(:manage, Project) { |p| p.something? }
```

My proposal is to add `:if` and `:unless` keys as special hash arguments to allow rule like these ones:

``` ruby
can :manage, Project, if: :something?
can :manage, Project, unless: :something_else?
```

This kind of conditions will make `Rule` class cacheable without need of block arguments, and will give more expressiveness for some use cases. 

The `Rule` class will become something like

``` ruby
def initialize(base_behavior, action, subject, conditions, block)
    raise Error, "You are not able to supply a block with a hash of conditions in #{action} #{subject} ability. Use either one." if conditions.kind_of?(Hash) && !block.nil?
  @match_all = action.nil? && subject.nil?
  @base_behavior = base_behavior
  @actions = [action].flatten
  @subjects = [subject].flatten
  @conditions = conditions || {}
  @if_method = conditions.delete(:if)
  @unless_method = conditions.delete(:unless)
  @block = block
end
# when checking conditions
subject.send(@if_method) if @if_method
```

What do you think? Are you interested in that? If so I can try to implement this feature in a feature branch.

PR #120, opened 2014-09-06
if/unless options for rules

fabn:
As requested in #105 here's my attempt to implement that.

See #105 for discussion and usage.

### DIFF
--- a/lib/cancan/rule.rb
+++ b/lib/cancan/rule.rb
@@ -34,7 +34,7 @@ module CanCan
     def matches_conditions?(action, subject, extra_args)
       if @match_all
         call_block_with_all(action, subject, extra_args)
-      elsif subject_conditions?
+      elsif subject_conditions? && !subject_class?(subject)
         match_subject_conditions?(subject)
       elsif @block && !subject_class?(subject)
         @block.call(subject, *extra_args)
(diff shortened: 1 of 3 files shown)


fabn:
@bryanrite any hope to get this merged? Thanks.

bryanrite:
Hi @fabn 

Sorry for taking so long on this, been busy :confused: 

I like the feature, the only reason I haven't merged it yet, is that it, along with a couple other PRs, like #192, are attempting to add context to rules but in a way that is not consistent with the way that other rules are added. Your additions would not work (and shouldn't) for collection style routes, but is written like a scope that could and I don't think that is obvious for some... and I would like the rules for an authorization framework to be unambiguous.

This is of course not your fault, but a limitation of the `ability` file, and you've implemented it really the only way you can.  I think we should have the ability to do `if`, `unless`, and `not`, along with controlling parameters sanitization, etc. so I am trying to redesign a DSL for the ability file to be unambiguous as well as more flexible when it comes to defining rules... which would include the functionality you've written here.

fabn:
@bryanrite thanks for the status update.

Senjai:
Hi @fabn,

I think this takes a good step towards the concept of using scopes instead of hash conditions, but I don't think the value is worth the complexity at the moment.

Currently, users can wrap can declarations in if blocks to solve the same problem and we don't have to support any additional code for this and we don't have to add an additional step in the matches_conditions method which is currently feeling some pain.

PR #127, opened 2014-09-29
Allow Rule to carry around a hash of options.

ejholmes:
This does 3 things.

* Extracts a private `match` method for more reusability.
* Allows a `CanCan::Rule` to carry a hash of options around.
* When a match is found but it doesn't allow access, that match is also provided to CanCan::AccessDenied.

The idea behind this would be to allow an api to respond with a reason for why the request is denied, so this allows you to define a rule like:

```ruby
cannot :join, Thing, nil, reason: :more_info
```

### DIFF
--- a/lib/cancan/ability.rb
+++ b/lib/cancan/ability.rb
@@ -126,8 +126,8 @@ module CanCan
     #     # check the database and return true/false
     #   end
     #
-    def can(action = nil, subject = nil, conditions = nil, &block)
-      rules << Rule.new(true, action, subject, conditions, block)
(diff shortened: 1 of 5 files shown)


Senjai:
Hi @ejholmes, Thank you for the submission. I think this is meant to address issues like #202 and #173 right? We might want to change a few things around how this is implemented, but I wouldn't oppose having metadata for rules. 

If the only reason is for messages, it might be better implemented using I18n. Or just for reporting args #173 might be the best approach in the short term.

PR #132, opened 2014-10-16
Mongoid Adapter: Nested hash conditions

vighnesh1987:
`accessible_by` does not work well with abilities that contain nested hash conditions on associations

If you have models related as: A belongs_to B belongs_to C and you want to define abilities on A with a hash condition that concerns fields in C.

Prior to this commit, can? would work with such nested conditions but accessible_by would not.

This fixes accessible_by.


Senjai:
@vighnesh1987 Thank you for your submission. Am I correct that this is the same issue as #156

Senjai:
@vighnesh1987 Thanks!

PR #135, opened 2014-10-22
Do not end initialization process before assign attributes

bronislav:
Make sure that `after_initialize` callback for resource will be called
after assigning attributes from abilities. This is possible because
most of AR methods are able to take optional block argument.

### DIFF
--- a/lib/cancan/controller_resource.rb
+++ b/lib/cancan/controller_resource.rb
@@ -77,8 +77,9 @@ module CanCan
     end
 
     def build_resource
-      resource = resource_base.new(resource_params || {})
-      assign_attributes(resource)
+      resource_base.new(resource_params || {}) do |resource|
+        assign_attributes(resource)
+      end
     end
 
     def assign_attributes(resource)
--- a/spec/cancan/controller_resource_spec.rb
+++ b/spec/cancan/controller_resource_spec.rb
@@ -14,6 +14,7 @@ describe CanCan::ControllerResource do
         attributes.each do |attribute, value|
           send("#{attribute}=", value)
         end
+        yield self if block_given?
       end
     end
 


benlangfeld:
@bronislav As far as I can see, there are still no tests here. You should add some if you want this to be merged :)

Senjai:
I don't necessarily see a problem with this, but I also don't see a reason for it. Is there a scenario where the current logic is failing or is this just an opinionated change?

bronislav:
My use case it when I assign default values for some entity depending on the the attribute assigned by cancan.

For example, I have company which have default time zone field. And when I create for example project entity I want copy default time zone from the company. This usually done in `after_initialize` callback. But in current implementation company field is `nil` because cancan didn't assign it yet. I can't move assignment from the `after_initialize` because I want show it on the new from fro user to be able to change default values.

Senjai:
@bronislav That's fair. I would highly recommend against using after_initialize :), most stuff like that typically happens on before_validation. 

The other issue I'd like you to address then is the fact that resource_base may not be an `ActiveRecord` object. Have you tested this with other adapters? If they cannot play nicely I would be in favor of keeping the original behavior instead of adding complexity to handle an edge case like this, as I believe very few people use after_initialize for this kind of logic.

bronislav:
If set default value in `before_validation` callback, then is won't appear on the new form. I know that using `after_initialize` is quite tricky but didn't find other solution.

About other adapters. When I made pull request I have checked all three of supported adapters to make sure that they may consume block during initialization process. So, I read their sources but didn't check it manually.

Senjai:
@bronislav With Datamapper alone it does not share the same behaviour as active record:

```ruby
[1] pry(CanCan::ModelAdapters::DataMapperAdapter)> model_class.create do |thing|
[1] pry(CanCan::ModelAdapters::DataMapperAdapter)*   thing.secret = true
[1] pry(CanCan::ModelAdapters::DataMapperAdapter)* end  
=> #<Article @id=7 @published=false @secret=false @priority=<not loaded>>
```

Along with that case, there's the issue where you want your attributes to be assigned before your after_initialize logic, but others may not want this. 

Others may want their after initialize logic to occur before assignment and the best way to solve this problem for both scenarios is to avoid using after_initialize. There's almost always a better way of solving the issue. 

Thank you for raising the issue.

PR #159, opened 2015-01-07
For ruby 1.8 and 1.9, we need a supported i18n version.

bryanrite:


bryanrite:
This unfortunately restricts i18n unnecessarily due to an issue with Bundler. While I'm waiting clarification on that issue, I'll merge this anyways since we're broken without it.

Bundler reference: https://github.com/bundler/bundler/issues/3340

bryanrite:
Forgot that Rails 4 needs 1.9.3 anyways so the issue is a bit moot as we are only restricting i18n to `< 0.7` for Rails 3 installs, which shouldn't cause anyone any headaches.

If the bundler issue is resolved then we're only restricting for Rails 3 installs on ruby < 1.9.3 which is what is actually required.

Issue #163, opened 2015-01-09
Postgres 'ProtocolViolation' error with overlapping abilities


fattymiller:
I have just updated to `Ruby 2.2.0` and `Rails 4.2` for this application and am getting the following in Postgres when using `cancancan`. 

```
PG::ProtocolViolation: ERROR:  bind message supplies 0 parameters, but prepared statement "" requires 2
```

> **tl;dr**
> The issue seems to be that the SQL parameters aren't passed in with the query when two or more abilities overlap.

> ```
WHERE (("company_users"."acl" = $1 AND "company_users"."user_id" = $2) OR ("company_users"."acl" = $1 AND "company_users"."user_id" = $2))
```
> versus
> ```
WHERE "company_users"."acl" = $1 AND "company_users"."user_id" = $2  [["acl", "admin"], ["user_id", 1]]
```

With my ability.rb file setup with the following:
```
    can :manage, Company, company_users: { acl: "admin", user_id: user.id }
    can :index, Company, company_users: { acl: "manager", user_id: user.id }
```
I get the error below, but if either one of the ability declarations are commented out, or they are changed to be exclusive of each other (example: `:admin_test` and `:manager_test`) the code works as expected.

When calling `Company.accessible_by(User.first.ability)` I have observed the following:

**Using `cancan` (`1.6.10`):**
```
[1] pry(main)> Company.accessible_by(User.first.ability)
  User Load (1.3ms)  SELECT  "users".* FROM "users"  ORDER BY "users"."id" ASC LIMIT 1
DEPRECATION WARNING: sanitize_sql_hash_for_conditions is deprecated, and will be removed in Rails 5.0
. (called from <main> at (pry):1)
DEPRECATION WARNING: sanitize_sql_hash_for_conditions is deprecated, and will be removed in Rails 5.0
. (called from <main> at (pry):1)
  Company Load (1.7ms)  SELECT "companies".* FROM "companies" INNER JOIN "company_users" ON "company_users"."company_id" = "companies"."id" WHERE (("company_users"."acl" = 'manager' AND "company_users"."user_id" = 1) OR ("company_users"."acl" = 'admin' AND "company_users"."user_id" = 1))
=> [#<Company:0x007fcb71d00fe0 id: 1, name: "Company 1", created_at: Fri, 09 Jan 2015 04:43:34 UTC +00:00, updated_at: Fri, 09 Jan 2015 04:43:34 UTC +00:00>]
```

[...]

PR #168, opened 2015-01-13
Fix Postgres 'ProtocolViolation' error with overlapping abilities

arambert:
fixing https://github.com/CanCanCommunity/cancancan/issues/163 using sanitize_sql_hash_for_conditions implementation in https://github.com/rails/rails/blob/4-2-stable/activerecord/lib/active_record/sanitization.rb


bryanrite:
This looks great, I'll try and get it merged and push ASAP.

Thank you!

PR #170, opened 2015-01-21
Updated travis.yml with ruby 2.2

sivagollapalli:


bryanrite:
Thank you @sivagollapalli 

The failure is due to a Ruby 2.2 regression (https://bugs.ruby-lang.org/issues/10314)... I'll try and get that fixed up and merged in asap.

bryanrite:
Following the Rails discussion (https://github.com/rails/rails/pull/18160), 3.2 will eventually be updated to work with this, but 3.1 will never be.

Do you mind adding 3.1 to the exclude list in travis and 3.2 to the allowed failures in this PR?

Thanks so much!

Issue #60, opened 2014-05-14
accessible_by Ignores Explicit ActiveRecord select()


bmulholland:
I'm using cancancan to get an index of objects from my database. In this case, each object has a very large text field (in some cases >1MB) that I don't need for the index option. I have found that this means that doing the index for some customers right now takes >7s!

Since all I need is a selection of info about each item in the database (e.g. title), I wanted to select only e.g. the title and id. However, since cancancan uses `includes` (not `joins`) internally, the select is completely ignored and there's no way to NOT select that large text field.

May I propose that, when using accessible_by, cancancan should use joins instead such that those using the gem can ignore large fields that may slow things down?

I believe the relevant line is here: https://github.com/CanCanCommunity/cancancan/blob/develop/lib/cancan/model_adapters/active_record_adapter.rb#L104

I'm happy to provide a PR if I can get a hint about how I might know at that point of the code how we're doing an `accessible_by` instead of a normal permissions check.

PR #171, opened 2015-01-21
Allow disabling ActiveRecord eager loading

bmulholland:
While #includes is useful to help avoid N+1 query problems, it can
cause challenges in other ways. In particular, complex queries that
involve large objects at the root of their chain can end up including
that large object for every row of the selected model.

As an example, imagine a blog post with many comments, each with an
author. We'd like to restrict access to the authors that have comments
on blog posts visible to the user. Your permission might look like:

```
can :read, User, {
  comment: { blog: { accessible_blogs: { user_id: @current_user.id } } }
}
```

Then you can do User.accessible_by(current_ability). That query will
give you a list of all the authors, but will also include the text of
the blog and comment. That becomes an issue for large blog posts with
many comments.

Using joins instead of includes, in this case, runs a query that only
selects User attributes.

Issue https://github.com/CanCanCommunity/cancancan/issues/60 has some more information on an application of this
fix.

---

Here's my first stab at allowing a toggle of includes/joins.

It sucks to have eager_load be passed to _all_ model_adapters, even if they
don't support this, but I can't think of a nicer way to do this. The argument
location seems poor, as well - is there a good alternative to passing it before
`where_conditions`?

### DIFF
--- a/CHANGELOG.rdoc
+++ b/CHANGELOG.rdoc
@@ -1,5 +1,7 @@
 Develop
 
+* Fix cancancan#60 - Allow disabling eager loading via ActiveRecord#includes
+
 
 1.10.1 (January 13th, 2015)
 
--- a/lib/cancan/model_adapters/abstract_adapter.rb
+++ b/lib/cancan/model_adapters/abstract_adapter.rb
@@ -47,7 +47,7 @@ module CanCan
         @rules = rules
       end
 
-      def database_records
+      def database_records(eager_load)
         # This should be overridden in a subclass to return records which match @rules
         raise NotImplemented, "This model adapter does not support fetching records from the database."
       end
--- a/lib/cancan/model_adapters/active_record_3_adapter.rb
+++ b/lib/cancan/model_adapters/active_record_3_adapter.rb
@@ -39,8 +39,12 @@ module CanCan
 
       private
 
-      def build_relation(*where_conditions)
-        @model_class.where(*where_conditions).includes(joins)
+      def build_relation(eager_load = true, *where_conditions)
+        if eager_load
+          @model_class.where(*where_conditions).includes(joins)
+        else
+          @model_class.where(*where_conditions).joins(joins)
+        end
       end
     end
   end
--- a/lib/cancan/model_adapters/active_record_4_adapter.rb
+++ b/lib/cancan/model_adapters/active_record_4_adapter.rb
@@ -12,9 +12,15 @@ module CanCan
       # look inside the where clause to decide to outer join tables
       # you're using in the where. Instead, `references()` is required
       # in addition to `includes()` to force the outer join.
-      def build_relation(*where_conditions)
+      def build_relation(eager_load = true, *where_conditions)
         relation = @model_class.where(*where_conditions)
-        relation = relation.includes(joins).references(joins) if joins.present?
+        if joins.present?
+          if eager_load
+            relation = relation.includes(joins).references(joins)
+          else
+            relation = relation.joins(joins)
+          end
+        end
         relation
       end
 
--- a/lib/cancan/model_adapters/active_record_adapter.rb
+++ b/lib/cancan/model_adapters/active_record_adapter.rb
@@ -60,14 +60,14 @@ module CanCan
         clean_joins(joins_hash) unless joins_hash.empty?
       end
 
-      def database_records
+      def database_records(eager_load = true)
         if override_scope
           @model_class.where(nil).merge(override_scope)
         elsif @model_class.respond_to?(:where) && @model_class.respond_to?(:joins)
           if mergeable_conditions?
-            build_relation(conditions)
+            build_relation(eager_load, conditions)
           else
-            build_relation(*(@rules.map(&:conditions)))
+            build_relation(eager_load, *(@rules.map(&:conditions)))
           end
         else
           @model_class.all(:conditions => conditions, :joins => joins)
--- a/lib/cancan/model_adapters/data_mapper_adapter.rb
+++ b/lib/cancan/model_adapters/data_mapper_adapter.rb
@@ -18,7 +18,7 @@ module CanCan
         !!collection.first(conditions)
       end
 
-      def database_records
+      def database_records(eager_load = true)
         scope = @model_class.all(:conditions => ["0 = 1"])
         cans, cannots = @rules.partition { |r| r.base_behavior }
         return scope if cans.empty?
--- a/lib/cancan/model_adapters/mongoid_adapter.rb
+++ b/lib/cancan/model_adapters/mongoid_adapter.rb
@@ -22,7 +22,7 @@ module CanCan
         subject.matches?( subject.class.where(conditions).selector )
       end
 
-      def database_records
+      def database_records(eager_load = true)
         if @rules.size == 0
           @model_class.where(:_id => {'$exists' => false, '$type' => 7}) # return no records in Mongoid
         elsif @rules.size == 1 && @rules[0].conditions.is_a?(Mongoid::Criteria)
--- a/lib/cancan/model_adapters/sequel_adapter.rb
+++ b/lib/cancan/model_adapters/sequel_adapter.rb
@@ -28,7 +28,7 @@ module CanCan
         end
       end
 
-      def database_records
+      def database_records(eager_load = true)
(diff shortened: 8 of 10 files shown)


bmulholland:
This was actually created to fix another problem we encountered, which isn't mentioned in #60. We were loading ~10k objects and each object was including a large (~1MB) text column. Doing so spiked our unicorn worker memory usage up to 1GB each, just from using cancan instead of doing it directly.

bmulholland:
Re: failing tests. I think that Activerecord 3.0 only works with the `instance_variable_get` I got, but 3.1+ needs `loaded?`. Is there a way to use different checks on different versions?

Senjai:
@bmulholland Thanks for the submission. You can compare active record versions, but we may be dropping support for 3.x entirely in the next major, but that hasn't been decided yet.

coorasse:
Why cancancan is taking care of preloading my associations?
This is causing me really a lot of issues. @bmulholland  PR is not only necessary, but `joins` should be the default behaviour in my opinion.

coorasse:
I think we have an issue here. from my point of view `includes` has been introduce to avoid problems with INNER JOINS. In particular, when using `joins` instead of `includes` this test will fail: https://github.com/CanCanCommunity/cancancan/blob/develop/spec/cancan/model_adapters/active_record_adapter_spec.rb#L157 

But `includes` does not only use OUTER JOINS instead of INNER JOINS but also preload the associations....and that's BAD. We should not preload associations or it will lead to performance issues.
By instance, I have the following structure:

```
class User
	has_many :group_participations
	has_many :groups, through: :group_participations
end

class GroupParticipation
	belongs_to :user
	belongs_to :group
end

class Group
	has_many :group_participations
	has_many :users, through: :group_participations
end

class Forum
	belongs_to :group
end

class Topic
	belongs_to :forum
end
```

and my visibility rule is

```
can :read, Topic, forum: {visible_outside: true, category: {visible_outside: true}}
can :read, Topic, forum: {group: {group_participations: {user_id: user.id}}}
```

which means that a user can see only the topics of the groups in which he is participating.

Ok, now in the controller I have

```
@last_topics = @group.topics.accessible_by(Ability.new(current_user)).order('topics.created_at desc').limit(10)
```

to retrieve the last 10 topics the user can see.

This leads to the following query with the current version of cancan:
```
SELECT  DISTINCT "topics"."id", topics.created_at AS alias_0 FROM "topics" 
LEFT OUTER JOIN "forums" "forums_topics" ON "forums_topics"."id" = "topics"."forum_id" 
LEFT OUTER JOIN "groups" ON "groups"."id" = "forums_topics"."group_id" 
LEFT OUTER JOIN "group_participations" ON "group_participations"."group_id" = "groups"."id" 
LEFT OUTER JOIN "categories" ON "categories"."id" = "forums_topics"."category_id" 
INNER JOIN "forums" ON "topics"."forum_id" = "forums"."id" 
WHERE "forums"."group_id" = $1 AND (("group_participations"."user_id" = 306) 
OR ("forums"."visible_outside" = 't' AND "categories"."visible_outside" = 't'))  
ORDER BY topics.created_at desc 
LIMIT 10
```

and immediatly after this one:

```
SELECT "topics"."id" AS t0_r0, "topics"."forum_id" AS t0_r1, "topics"."user_id" AS t0_r2, "topics"."subject" AS t0_r3, "topics"."locked" AS t0_r4, "topics"."pinned" AS t0_r5, "topics"."hidden" AS t0_r6, "topics"."state" AS t0_r7, "topics"."last_post_at" AS t0_r8, "topics"."views_count" AS t0_r9, "topics"."slug" AS t0_r10, "topics"."created_at" AS t0_r11, "topics"."updated_at" AS t0_r12, "topics"."token" AS t0_r13, "forums_topics"."id" AS t1_r0, 
"forums_topics"."name" AS t1_r1, "forums_topics"."description" AS t1_r2, "forums_topics"."category_id" AS t1_r3, "forums_topics"."group_id" AS t1_r4, "forums_topics"."views_count" AS t1_r5, "forums_topics"."slug" AS t1_r6, "forums_topics"."visible_outside" AS t1_r7, 
"groups"."id" AS t2_r0, "groups"."name" AS t2_r1, "groups"."description" AS t2_r2, "groups"."accept_requests" AS t2_r3, "groups"."interest_border_id" AS t2_r4, "groups"."facebook_page_url" AS t2_r5, "groups"."image_id" AS t2_r6, "groups"."title_bar" AS t2_r7, "groups"."old_image_url" AS t2_r8, "groups"."participation_role_id" AS t2_r9, "groups"."created_at" AS t2_r10, "groups"."updated_at" AS t2_r11, "groups"."change_advanced_options" AS t2_r12, "groups"."default_anonima" AS t2_r13, "groups"."default_visible_outside" AS t2_r14, "groups"."default_secret_vote" AS t2_r15, "groups"."max_storage_size" AS t2_r16, "groups"."actual_storage_size" AS t2_r17, "groups"."enable_areas" AS t2_r18, "groups"."group_participations_count" AS t2_r19, "groups"."image_file_name" AS t2_r20, "groups"."image_content_type" AS t2_r21, "groups"."image_file_size" AS t2_r22, "groups"."image_updated_at" AS t2_r23, "groups"."admin_title" AS t2_r24, "groups"."private" AS t2_r25, "groups"."rule_book" AS t2_r26, "groups"."subdomain" AS t2_r27, "groups"."certified" AS t2_r28, "groups"."status" AS t2_r29, "groups"."status_changed_at" AS t2_r30, "groups"."slug" AS t2_r31, "groups"."disable_participation_requests" AS t2_r32, "groups"."disable_forums" AS t2_r33, "groups"."disable_documents" AS t2_r34, "groups"."proposals_count" AS t2_r35, "groups"."meeting_organizations_count" AS t2_r36, 
"group_participations"."id" AS t3_r0, "group_participations"."user_id" AS t3_r1, "group_participations"."group_id" AS t3_r2, "group_participations"."participation_role_id" AS t3_r3, "group_participations"."created_at" AS t3_r4, "group_participations"."updated_at" AS t3_r5, "group_participations"."acceptor_id" AS t3_r6, 
"categories"."id" AS t4_r0, "categories"."name" AS t4_r1, "categories"."slug" AS t4_r2, "categories"."group_id" AS t4_r3, "categories"."created_at" AS t4_r4, "categories"."updated_at" AS t4_r5, "categories"."visible_outside" AS t4_r6 FROM "topics" 
LEFT OUTER JOIN "forums" "forums_topics" ON "forums_topics"."id" = "topics"."forum_id" 
LEFT OUTER JOIN "groups" ON "groups"."id" = "forums_topics"."group_id" 
LEFT OUTER JOIN "group_participations" ON "group_participations"."group_id" = "groups"."id" LEFT OUTER JOIN "categories" ON "categories"."id" = "forums_topics"."category_id" 
INNER JOIN "forums" ON "topics"."forum_id" = "forums"."id" 
WHERE "forums"."group_id" = 185 AND (("group_participations"."user_id" = 306) OR ("forums"."visible_outside" = 't' AND "categories"."visible_outside" = 't')) 
AND "topics"."id" IN (2034, 2033, 2032, 2031, 2030, 2024, 2023, 2022, 1953, 1952) 
ORDER BY topics.created_at desc
```

which I actually don't really need and preloads a lot of unnecessary stuff taking a lot of time!


With standard JOINS this leads instead to the following:

```
SELECT  DISTINCT "topics".* 
FROM "topics" 
INNER JOIN "forums" "forums_frm_topics" ON "forums_frm_topics"."id" = "topics"."forum_id" 
INNER JOIN "groups" ON "groups"."id" = "forums_topics"."group_id" 
INNER JOIN "group_participations" ON "group_participations"."group_id" = "groups"."id" 
INNER JOIN "categories" ON "categories"."id" = "forums_topics"."category_id" 
INNER JOIN "forums" ON "topics"."forum_id" = "forums"."id" 
WHERE "forums"."group_id" = 185 AND 
(("group_participations"."user_id" = 306) OR ("forums"."visible_outside" = 't' AND "categories"."visible_outside" = 't'))  
ORDER BY topics.created_at desc LIMIT 10
```

which is ok for me in this specific situation but which can lead to problems because of INNER JOINS.

coorasse:
I think CanCanCan should never take care of eager loading associations. That's why I'll close this PR in favor of #238 . Feel free to reopen

PR #178, opened 2015-02-05
Significantly improve rule lookup time

amarshall:
CanCan(Can)’s rule lookup time is currently abysmally slow with a non-small number of rules. This patch makes it ~300x faster with a large number of rules (see §Methodology & §Results below). This is accomplished by changing the time complexity of rule lookup from O(n) to O(1).

## Methodology

100 subjects, 5 actions each. No complex rules or queries. For more detail, see the actual [benchmarking scripts](https://gist.github.com/amarshall/babd7b51087c4d7577a7).

## Results

Some summary statistics (times in µs):

#### Before
<pre>
Read 3200 items
   Min. 1st Qu.  Median    Mean 3rd Qu.    Max.
   8039   15130   16760   16190   19630   45760
</pre>

#### After
<pre>
Read 3200 items
   Min. 1st Qu.  Median    Mean 3rd Qu.    Max.
  36.00   40.00   44.00   50.76   56.00  374.00
</pre>

#### Lookup time, constant rule numbers

Box plot showing before/after comparison
*Note the x-axis is in log scale.
(reproduce using [benchmark script](https://gist.github.com/amarshall/babd7b51087c4d7577a7#file-performance-single-rb))

#### Growth comparison

Varying the number of subjects clearly shows the change from O(n) → O(1):

Line plot showing difference in time complexity with varying number of subjects
(reproduce using [benchmark script](https://gist.github.com/amarshall/babd7b51087c4d7577a7#file-performance-growth-rb))

It should be clear now these changes bring a significant performance benefit. On a large application (see §Testing for details about size). We’ve seen pages where CanCan(Can) consumed as much as 25% of the page load time be reduced to ~3%.

## Testing performed

Since this is no-doubt a significant change, these changes went through testing using the existing test suite of a relatively large application (~100k Ruby SLOC) with ~400 CanCan rules spanning ~1000 lines. Some issues were found using this, in particular the rule-ordering problem, for which a test has been added to CanCanCan. This is, of course, in addition to CanCanCan’s own test suite.

### DIFF
--- a/lib/cancan/ability.rb
+++ b/lib/cancan/ability.rb
@@ -133,7 +133,7 @@ module CanCan
     #   end
     #
     def can(action = nil, subject = nil, conditions = nil, &block)
-      rules << Rule.new(true, action, subject, conditions, block)
+      add_rule(Rule.new(true, action, subject, conditions, block))
     end
 
     # Defines an ability which cannot be done. Accepts the same arguments as "can".
@@ -149,7 +149,7 @@ module CanCan
     #   end
     #
     def cannot(action = nil, subject = nil, conditions = nil, &block)
-      rules << Rule.new(false, action, subject, conditions, block)
+      add_rule(Rule.new(false, action, subject, conditions, block))
     end
 
     # Alias one or more actions into another one.
@@ -247,7 +247,7 @@ module CanCan
 
     def merge(ability)
       ability.rules.each do |rule|
-        rules << rule.dup
+        add_rule(rule.dup)
       end
       self
     end
@@ -255,7 +255,7 @@ module CanCan
     protected
 
     def rules
-      @rules ||= []
+      @rules.values.flatten
     end
 
     private
@@ -308,15 +308,38 @@ module CanCan
       results
     end
 
+    def add_rule(rule)
+      @rules ||= Hash.new { |h, k| h[k] = [] }
+      @rules[:all] << rule if rule.subjects.compact.empty?
+      rule.subjects.each do |subject|
+        @rules[subject] << rule
+      end
+    end
+
+    def alternative_subjects(subject)
+      subject = subject.class unless subject.is_a?(Module)
+      descendants = []
+      [:all, *subject.ancestors, *descendants, subject.class.to_s]
+    end
+
     # Returns an array of Rule instances which match the action and subject
     # This does not take into consideration any hash conditions or block statements
     def relevant_rules(action, subject)
-      relevant = rules.select do |rule|
+      return [] unless @rules
+      relevant = possible_relevant_rules(subject).select do |rule|
         rule.expanded_actions = expand_actions(rule.actions)
         rule.relevant? action, subject
       end
-      relevant.reverse!
+      relevant.reverse!.uniq!
       relevant
+    end
+
+    def possible_relevant_rules(subject)
+      if subject.is_a?(Hash)
+        rules
+      else
+        @rules.values_at(subject, *alternative_subjects(subject)).flatten
+      end
     end
 
     def relevant_rules_for_match(action, subject)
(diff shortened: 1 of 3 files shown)


bryanrite:
At first glance, this looks _really_ good. I've been meaning to get at something like this as well, so thanks so much for your work and detailed analysis. I'll take a more through look and see if I can get this merged in soon.

dgilperez:
:+1: :100:

Ch4s3:
The splat operator here:
````ruby
[:all, *subject.ancestors, *descendants, subject.class.to_s]
````

breaks 1.8.7 which is lacking some of that functionality. Should the gem even support 1.8.7? If so, should we put in a check in that uses the old method Ruby < 1.9?

**Comment by Ch4s3** [comment]
I put in a pull here https://github.com/amarshall/cancancan/pull/1 to try to fix this.

**Comment by amarshall** [comment]
Personally, I’m of the opinion that 1.8.7 support should be outright dropped as it hasn’t been maintained for some time now.

**Comment by Ch4s3** [comment]
@amarshall I totally agree, but I can't speak for the community. @bryanrite Would it be better to make it work for 1.8.7 just to get the feature merged or would you prefer to discuss future support for 1.8.7?

**Comment by dgilperez** [comment]
+1 to locking 1.8.7 support to the latest version, adding a note to the README and move forward.

**Comment by bryanrite** [comment]
Sorry I've been slow to respond, I've had a lot of things on my plate.

I'm looking at wrapping up a lot of breaking changes in the last few weeks into a new major release that would drop support for ruby < 2.0 and rails < 4.0. I think it would significantly simplify a lot of CanCan, but I'm still in the middle of planning it all out.

**Comment by Ch4s3** [comment]
@bryanrite I'm all for dropping support for Ruby < 2.0 and Rails < 4.0. Maybe it makes sense to put together a 2.0pre release that drops older versions and rolls in this branch and #34. What else is on the roadmap?

**Comment by bryanrite** [comment]
@Ch4s3 I'm putting together a working document that I'll circulate a bit with some ideas and goals. Right now, I'm thinking we'll have to jump to a 3.x, since Ryan started on a 2.x that was never officially released but does have some usage.

**Comment by Ch4s3** [comment]
@bryanrite that sounds great. Yeah, I was wondering about the whole Ryan/2.x situation.

**Comment by jackquack** [comment]
I love this branch. I did something similar once upon a time, but hacky: https://github.com/jackquack/cancan/commit/9f46adf19571e3b85ba932f55675c7a059560934

This PR looks much better though. I would love to get back to mainline.

**Comment by Ch4s3** [comment]
So where are we on this?

**Comment by Senjai** [comment]
@Ch4s3 Hi there, sorry about the delay. I'll add this to the docket for when 1.8 is dropped. That's a given.

**Comment by Ch4s3** [comment]
@Senjai awesome! thanks for the update.

File: lib/cancan/ability.rb [hunk 1]
```diff
@@ -302,21 +308,51 @@ def aliases_for_action(action)
...
+      subjects << :all if subjects.empty?
+
+      subjects.each do |subject|
+        @rules_index[subject] << position
+      end
+    end
+
+    def alternative_subjects(subject)
+      subject = subject.class unless subject.is_a?(Module)
+      descendants = []
```
jhawthorn: is this always `[]`? Why do we need this?

Senjai:
@Ch4s3 Sorry about the wait, I've been fairly busy lately (we all are :(). I rebased your changes and addressed jhawthorns feedback. Thank you -very- much for this contribution.

Senjai:
@amarshall ^ (mispinged someone else, sorry!)

jhawthorn:
:tada: this is extremely awesome. Thank you

Ch4s3:
Nice

PR #181, opened 2015-02-18
params[:controller] is never "camelized"

jhawthorn:
I don't believe that `params[:controller]` ever comes in as a camelized string (like `MyEngine::ModelsController`). In every version of rails I know it is underscored (like `my_engine/models`). Please correct me if I am wrong on this.

This provides a tiny simplification to the ControllerResource class. If this can be merged I intend to do some further cleanup (using controller_name, splitting ControllerResource into multiple classes)

### DIFF
--- a/lib/cancan/controller_resource.rb
+++ b/lib/cancan/controller_resource.rb
@@ -254,7 +254,7 @@ module CanCan
     end
 
     def namespace
-      @params[:controller].split(/::|\//)[0..-2]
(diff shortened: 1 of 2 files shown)


Senjai:
@jhawthorn :+1: I don't think it has been for a long time, and for sure it isn't now (https://github.com/rails/rails/blame/master/actionpack/lib/action_dispatch/routing/route_set.rb#L77)

PR #184, opened 2015-02-25
Fixes #139.

### LINKED ISSUE #139: load_and_authorize_resource doesn't assign attributes for update action
Even though the docs suggest that this is so. The strong parameters section even mentions that it checks for an update_params method.

kevintyll:
Used the tests by @rywall to fix attribute assignment bug in update action.

### DIFF
--- a/lib/cancan/controller_resource.rb
+++ b/lib/cancan/controller_resource.rb
@@ -114,7 +114,7 @@ module CanCan
     end
 
     def set_resource_params(resource)
-      Hash(resource_params).each do |attr, value|
+      (resource_params || {}).each do |attr, value|
         resource.send("#{attr}=", value)
       end
       resource
--- a/lib/cancan/controller_resource.rb
+++ b/lib/cancan/controller_resource.rb
@@ -60,7 +60,7 @@ module CanCan
       if !parent? && new_actions.include?(@params[:action].to_sym)
         build_resource
       elsif id_param || @options[:singleton]
-        find_resource
+        set_resource_params(find_resource)
       end
     end
 
@@ -111,6 +111,13 @@ module CanCan
           adapter.find(resource_base, id_param)
         end
       end
+    end
+
+    def set_resource_params(resource)
+      Hash(resource_params).each do |attr, value|
+        resource.send("#{attr}=", value)
+      end
+      resource
     end
 
     def adapter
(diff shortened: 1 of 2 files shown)


jhawthorn:
:-1: I believe this was working as intended. Your application code should be responsible for performing the update. `load_and_authorize_resource` should only load and authorize.

Where in the docs did it suggest otherwise? We should fix that.

rywall:
It suggests this in the Strong Params section of the docs: https://github.com/CanCanCommunity/cancancan#strong-parameters

It also works this way in the create action, further adding to the confusion.

kevintyll:
Correct, the authorization was happening on updates without loading the form data, so I still had to manually load and authorize the object manually in the update actions.  Since the docs stated it should use strong params to load the object before authorizing per the link rywall provided, I went ahead and implemented that functionality.

Senjai:
@rywall @kevintyll Thank you for your report but I agree with @jhawthorn that `find_resource` should never modify the resource it finds. The reason attributes are assigned in the create action is because it's considered a new action. 

The attributes are assigned in this case as a convenience as any attributes on the model at that time would be sourced from the form. When updating, unless you inspect the changed attributes, a found resource would appear to have values that would not be persisted along with others that may be persisted which can be problematic.

I assume that you have a situation like the following:

```ruby
  can :update, Thing, something: true
```

Which would fail authorization unless the value something is set to true before the authorization check is called. If you must handle this situation, I don't think CanCan should infer this for you. 

This is a hint that you might be able to architect your application a little more robustly. It makes no sense to allow this to happen when additional logic may eventually set it to something else. It only makes sense if `something: true` is persisted **before** the update is called.

rywall:
Maybe I'm missing something but the strong parameters section of the README clearly suggests it will assign the attributes...

> By default, CanCan will try to sanitize the input on :create and **:update** routes by seeing if your controller will respond to the following methods

Am I misinterpreting the README?

riffraff:
I second what @rywall the wiki page also says 

> If you specify a create_params or update_params method, CanCan will run that method depending on the action you are performing.

but `update_params` will in fact not be called in any way, AFAICT, so that probably also needs updating.

Senjai:
@riffraff Would you be willing to make a pull request to the fact?

riffraff:
not sure if this is ok, but if it is, I'll also update the wiki.

sintro:
The readme is really incorrect in update_params description section. Suggested in PR read is better anyway.

PR #186, opened 2015-03-03
    Extract resource-loading logic from ControllerResource into another class.

jcoglan:
I was reading the source of `ControllerResource` to try to understand why an `authorize_resource` macro in our controller was authorizing a different resource than the one we expected. I found it quite hard to see what was going on, because all the logic for `load_resource` is mixed into the same class as that for `authorize_resource`.

This commit extracts all the loading logic into a new class called `ControllerResource::Loader`, and refactors some internal APIs in the process to support the interaction between `Loader` and `ControllerResource`. These refactorings include:
- Extracing values that are a matter of definition (e.g. the options to `before_filter` and the lists of types of actions) into named constants.
- Pulling some repeated expressions into their own methods, e.g. `controller_action` and `collection_name`.
- Inlining some methods that are only called from one place, e.g. `initial_attributes`, `id_param_key`, `resource_class_with_parent`, `params_methods`, `namespace`.
- Eliminate some overly long expressions, and reduce complexity of a few methods by flattening conditional blocks.
- Using `__send__` rather than `send` to avoid accidentally calling user-defined domain methods called `send`.

### DIFF
--- a/lib/cancan.rb
+++ b/lib/cancan.rb
@@ -2,6 +2,7 @@ require "cancan/version"
 require 'cancan/ability'
 require 'cancan/rule'
 require 'cancan/controller_resource'
+require 'cancan/controller_resource/loader'
 require 'cancan/controller_additions'
 require 'cancan/model_additions'
 require 'cancan/exceptions'
--- a/lib/cancan/controller_resource.rb
+++ b/lib/cancan/controller_resource.rb
@@ -107,7 +107,7 @@ module CanCan
 
     def parent_tuple
       [*@options[:through]].each do |name|
-        parent = fetch_parent(name)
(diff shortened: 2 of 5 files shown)


jcoglan:
I'm not sure why all these merge commits are in this PR, I only intended for my own commit to be here.

jcoglan:
Never mind, I rebased it on top of `develop` and it looks okay now.

Yoshyn:
:+1:

coorasse:
please rebase on `feature/v2` branch

coorasse:
I think this was a nicely done work but we have too many conflicts with the current version.
A new refactoring of the current develop branch would be more then welcome. I'll close this PR. Thank you

PR #190, opened 2015-03-08
We must assign resource_instance before asking the controlller for resource_params

dima4p:
Sometimes (as in my case) the controller calculates the list of allowed fields in resource_params based on the current_user abilities. The latter needs the resource to be instantiated. But it turned out that it is not done for new_actions.
The commit fixes the situation.

### DIFF
--- a/lib/cancan/controller_resource.rb
+++ b/lib/cancan/controller_resource.rb
@@ -77,7 +77,8 @@ module CanCan
     end
 
     def build_resource
-      resource = resource_base.new(resource_params || {})
+      self.resource_instance = resource = resource_base.new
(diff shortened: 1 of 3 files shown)


Senjai:
@dima4p Can you please provide a regression spec demonstrating where this fails as well as the fix?

dima4p:
All the problems were in spec files that are fixed now. ;-)

Senjai:
@dima4p closing this, as I understand that this is no longer an issue right? Let me know if I'm mistaken.

dima4p:
Yes, you are mistaken. The bug remains as you did not merge the pull request.

Issue #191, opened 2015-03-15
What's the best way to write a negative condition?


marnen:
I'd like to put a negative condition in my Ability definition. What I'd ideally like is something like 
```
cannot :update, TimeEntry, not(invoice_id: nil)
```
or a Squeel hook so I could do
```
cannot :update, TimeEntry, { :invoice_id.not_eq => nil }
```

It's not practical to reverse the sense of the conditional here, because the `cannot` is being used to restrict a previous `can`. If I change it to
```
can :update, TimeEntry, invoice_id: nil
```
then it won't interact correctly with the previous broader `can`. Help?

bryanrite:
I would suggest using ability with a block. You can specify a scope and/or more control at the instance level: https://github.com/CanCanCommunity/cancancan/wiki/Defining-Abilities-with-Blocks#block-conditions-with-scopes

marnen:
@bryanrite I thought about a block, but my understanding was that it wouldn't translate into a direct database query, and so it would be far more inefficient than the hash form. Am I wrong about this?

FWIW, Rails 4 supports `where.not(conditions_hash)`, so ideally I'd love to see cancancan interface with that. Obviously, the problem is that we pass a hash straight to `where`, so there's no way to express `where.not`. Hmm. Maybe I'll figure out some sort of patch for that if there's no good way for the time being.

bryanrite:
It uses the scope part on collection routes so it is efficient for multiple records, and uses the block part for instance routes, where the instance is already loaded... so you shouldn't have any issues with efficiency. In fact, if you only specify a block and try to authorize a collection-type route, it'll complain.

I agree it would be nice to be able to use some of the new Rails 4 query functionality.

marnen:
That also means I have to write the same query twice, once in the scope definition and once as a block. I really, really don't like that.

I'm currently trying to patch `ActiveRecord4Adapter`.

marnen:
Pull request at #192.

PR #192, opened 2015-03-16
Negative conditions for ActiveRecord 4

marnen:
Support negative conditions (`where.not`) for ActiveRecord 4. Usage: `can :read, Model, not: {name: 'foo'}`. See #191 for discussion.

### DIFF
--- a/lib/cancan/model_adapters/active_record_4_adapter.rb
+++ b/lib/cancan/model_adapters/active_record_4_adapter.rb
@@ -2,8 +2,23 @@ module CanCan
   module ModelAdapters
     class ActiveRecord4Adapter < AbstractAdapter
       include ActiveRecordAdapter
-      def self.for_class?(model_class)
-        model_class <= ActiveRecord::Base
+
+      class << self
+        def for_class?(model_class)
+          model_class <= ActiveRecord::Base
+        end
+
+        def override_condition_matching?(subject, name, value)
+          name == :not || super
+        end
+
+        def matches_condition?(subject, name, value)
+          if name == :not && value.kind_of?(Hash)
+            value.none? {|attribute, test| subject.send(attribute) == test }
+          else
+            super
+          end
+        end
       end
 
       private
--- a/lib/cancan/model_adapters/active_record_4_adapter.rb
+++ b/lib/cancan/model_adapters/active_record_4_adapter.rb
@@ -8,14 +8,32 @@ module CanCan
 
       private
 
+      def association_condition?(name, value)
+        name != :not && super
+      end
+
       # As of rails 4, `includes()` no longer causes active record to
       # look inside the where clause to decide to outer join tables
       # you're using in the where. Instead, `references()` is required
       # in addition to `includes()` to force the outer join.
       def build_relation(*where_conditions)
+        if where_conditions.kind_of?(Array) && where_conditions.size == 1
+          condition = where_conditions.first
+          if condition.kind_of? Hash
+            not_conditions = condition.delete(:not)
+            where_conditions = [condition]
+          end
+        end
+
         relation = @model_class.where(*where_conditions)
+        relation = relation.where.not(not_conditions) if not_conditions
         relation = relation.includes(joins).references(joins) if joins.present?
         relation
+      end
+
+      def merge_joins(base, add)
+        add.delete :not
+        super base, add
       end
 
       # Rails 4.2 deprecates `sanitize_sql_hash_for_conditions`
--- a/lib/cancan/model_adapters/active_record_adapter.rb
+++ b/lib/cancan/model_adapters/active_record_adapter.rb
@@ -30,7 +30,7 @@ module CanCan
       def tableized_conditions(conditions, model_class = @model_class)
         return conditions unless conditions.kind_of? Hash
         conditions.inject({}) do |result_hash, (name, value)|
-          if value.kind_of? Hash
+          if association_condition? name, value
             value = value.dup
             association_class = model_class.reflect_on_association(name).klass.name.constantize
             nested = value.inject({}) do |nested,(k,v)|
@@ -75,6 +75,10 @@ module CanCan
       end
 
       private
+
+      def association_condition?(name, value)
+        value.kind_of? Hash
+      end
 
       def mergeable_conditions?
         @rules.find {|rule| rule.unmergeable? }.blank?
--- a/spec/cancan/model_adapters/active_record_4_adapter_spec.rb
+++ b/spec/cancan/model_adapters/active_record_4_adapter_spec.rb
@@ -6,36 +6,40 @@ if defined? CanCan::ModelAdapters::ActiveRecord4Adapter
       before :each do
         ActiveRecord::Base.establish_connection(:adapter => "sqlite3", :database => ":memory:")
         ActiveRecord::Migration.verbose = false
-        ActiveRecord::Schema.define do
-          create_table(:parents) do |t|
-            t.timestamps :null => false
-          end
-
-          create_table(:children) do |t|
-            t.timestamps :null => false
-            t.integer :parent_id
-          end
-        end
-
-        class Parent < ActiveRecord::Base
-          has_many :children, lambda { order(:id => :desc) }
-        end
-
-        class Child < ActiveRecord::Base
-          belongs_to :parent
-        end
-
         (@ability = double).extend(CanCan::Ability)
       end
 
-      it "respects scope on included associations" do
-        @ability.can :read, [Parent, Child]
+      context 'associations' do
+        before :each do
+          ActiveRecord::Schema.define do
+            create_table(:parents) do |t|
+              t.timestamps :null => false
+            end
 
-        parent = Parent.create!
-        child1 = Child.create!(:parent => parent, :created_at => 1.hours.ago)
-        child2 = Child.create!(:parent => parent, :created_at => 2.hours.ago)
+            create_table(:children) do |t|
+              t.timestamps :null => false
+              t.integer :parent_id
+            end
+          end
 
-        expect(Parent.accessible_by(@ability).order(:created_at => :asc).includes(:children).first.children).to eq [child2, child1]
+          class Parent < ActiveRecord::Base
+            has_many :children, lambda { order(:id => :desc) }
+          end
+
+          class Child < ActiveRecord::Base
+            belongs_to :parent
+          end
+        end
+
+        it "respects scope on included associations" do
+          @ability.can :read, [Parent, Child]
+
+          parent = Parent.create!
+          child1 = Child.create!(:parent => parent, :created_at => 1.hours.ago)
+          child2 = Child.create!(:parent => parent, :created_at => 2.hours.ago)
+
+          expect(Parent.accessible_by(@ability).order(:created_at => :asc).includes(:children).first.children).to eq [child2, child1]
+        end
       end
     end
 
--- a/spec/cancan/model_adapters/active_record_4_adapter_spec.rb
+++ b/spec/cancan/model_adapters/active_record_4_adapter_spec.rb
@@ -41,6 +41,31 @@ if defined? CanCan::ModelAdapters::ActiveRecord4Adapter
           expect(Parent.accessible_by(@ability).order(:created_at => :asc).includes(:children).first.children).to eq [child2, child1]
         end
       end
+
+      context 'conditions' do
+        before :each do
+          ActiveRecord::Schema.define do
+            create_table(:records) do |t|
+              t.timestamps :null => false
+              t.string :name
+            end
+          end
+
+          class Record < ActiveRecord::Base
+          end
+        end
+
+        it 'uses :not for negative conditions' do
+          @ability.can :read, Record, :not => {:name => 'unreadable'}
+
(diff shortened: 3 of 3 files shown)


marnen:
I note that in two weeks, there's been no response on this. Is there something more that you're waiting for?

bryanrite:
Hi @marnen 

I've been trying to put together a document for a major Cancancan update (3.0) which would incorporate changes like this and others into a more refined syntax, and something that leverages AR4 better.  I've been slow to add this because I do not want to merge this in and have it change immediately.

marnen:
I get that, but better that we have something that works today than nothing at all.

Senjai:
@marnen Hi Marnen, this is definitely on the list of things we want to accomplish. There are a few outstanding issues around the adapters as is that I'd like to take a look at before bringing new features like this in. Apologies for the wait.

As a workaround in the meantime, you should be able to accomplish this with a block.

Senjai:
@marnen Are you still wanting this in? I'm hesitant to merge this, as I might misunderstand how it's exactly supposed to be used.

I hacked up a quick spec to experiment with this: https://github.com/Senjai/cancancan/commit/bcce8d296b9bbdfbd6c6fd1a4f234d940feda3d8

I would expect one of those specs to pass. As I would expect users might assume that this would work correctly. (our hash syntax is already pretty confusing)

Am I just misunderstanding how this should be used? Or do we need to cover additional cases?

coorasse:
Isn't the example provided already covered by scopes?
You can either define that:
```ruby
can :read, Model, ['name != ?', 'foo'] do |model|
  model.name != 'foo'
end
```

or that:
```ruby
can :read, Model, Model.no_name('foo') do |model|
  !model.has_name('foo')
end

class Model
 scope :no_name, ->(prop) { where.not(name: prop) }

 def has_name?(prop)
   name != prop
 end
end
```

marnen:
@Senjai Yes, I still think this is a feature we should have. I'm not sure why you're hesitant about this; see detailed discussion in #191. However, if @coorasse is correct about scopes and/or `where` parameters, then it's less necessary. I do still think, though, that supporting `:not` as in my pull request is valuable: with @coorasse's suggestion, I'd potentially have to make a negative and a positive version of each scope, and I'd like to avoid that.

Senjai:
The problem is that the interface isn't consistent with respect to the test here: https://github.com/Senjai/cancancan/commit/bcce8d296b9bbdfbd6c6fd1a4f234d940feda3d8

It also isn't consistent with how the hash syntax has been translated to active record. Up until now, any were condition used with active record could be translated pretty easily to the hash condition for cancan.

But `Spree::Order.where(not: { number: "1234" })` is not valid, for example.

In these scenarios in the past I have done:

`cannot :update, Thing, roles: { name: "cannot-update-things" }`
which would stand in place of:
`can :update, Thing, not: { roles: { name: "cannot-update-things" } }`


I also use scopes for complex situations, if they exist `cannot :update, Thing, Thing.supercomplexquery`

Especially if raw SQL is required. With a general authorization flow, your authorization should not be infinitely complex. If it is, you're going to want to use a scope anyway instead of a multi-line hash.

coorasse:
Starting from your first example, what wouldn't work with
`cannot :read, Model, name: 'foo'`?

marnen:
@coorasse: Have you actually read #191 as mentioned in the initial pull request description?

coorasse:
@marnen I read it but without a real, concrete, example is hard for me to understand your issue.
In #191 you write that `can :update, TimeEntry, invoice_id: nil` `won't interact correctly with the previous broader can`. 
Can I ask you to show me a concrete scenario? 
Why do you have a broader can if you cannot actually update elements which already have an id? Can't you just restrict the first `can?`?

marnen:
@coorasse: Sure. The example I used in #191 is taken from https://github.com/marnen/duckbill/blob/007c7ae4a26278222bbe3420b0c2557d24d6f3c7/app/models/ability.rb#L36-L37, where users can manage time entries on their own clients' projects, *unless* those time entries have already been invoiced. So I used my forked version of cancancan (as seen in this pull request) and wrote
```ruby
can :manage, TimeEntry, project: {client: {user_id: user.id}}
cannot [:destroy, :update], TimeEntry, not: {invoice_id: nil}
```

Note carefully, that in this case, it would not have worked if the second line had been 
```ruby
can [:destroy, :update], TimeEntry, invoice_id: nil
```
(which was actually what I tried first). This is because the `cannot` is *taking away* permissions granted by the previous *can*, and (so far as I can tell) a `can` can only *add* permissions.

marnen:
@Senjai:
Something I initially missed from your last comment:
> A negative and positive version would not be required, because you can simply:
```ruby
cannot :update, Thing, Thing.scopestuff
or can :update, Thing, Thing.scopestuff
```

I think you misunderstood the use case I was talking about. Since `can P` and `cannot !P` are not equivalent in the way they interact with other rules, I may need to write `can :update, Thing, Thing.not_scopestuff`, even if I already have a `Thing.scopestuff` scope. That's what I meant about having to create a negative and positive version.

In the example here, assuming I already have `Thing.scopestuff`, it makes no sense to define `Thing.not_scopestuff`. If I need the `can` form, I should just be able to do `can :update, Thing, not: Thing.scopestuff`.

coorasse:
I thing you should follow a different approach when defining your permissions.
Cancancan starts giving the minimum amount of permissions (nobody can do anything) and suggests that you give "increasing" permissions to your users.
What you are doing here is giving first larger permissions, and then you try to restrict them again.
From my point of view this approach is not wrong, but very hard to manage and counterintuitive.
Please follow my suggestion: instead of this:
```ruby
can :manage, TimeEntry, project: {client: {user_id: user.id}}
cannot [:destroy, :update], TimeEntry, not: {invoice_id: nil}
```

try to give first "smaller" permissions and then larger.

```ruby
can [:destroy, :update], TimeEntry, invoice_id: nil, project: {client: {user_id: user.id}}
can :read, TimeEntry, project: {client: {user_id: user.id}}
```

You should redefine your sentence: `users can manage time entries on their own clients' projects, unless those time entries have already been invoiced` into
`users can update and destroy time entries of their own client's project if time entries are not already invoiced. users can always read time entries of their own client's project'

Doing that you give larger permissions step by step and you don't need `cannot` at all.

marnen:
@coorasse:
> What you are doing here is giving first larger permissions, and then you try to restrict them again.

Yes, that is exactly what I want to be doing, kind of like `order Allow,Deny` in an `.htaccess` file. I want to say "I can manage a TimeEntry on my own projects *unless* the following conditions apply".

> try to give first "smaller" permissions and then larger.
```ruby
can [:destroy, :update], TimeEntry, invoice_id: nil, project: {client: {user_id: user.id}}
can :read, TimeEntry, project: {client: {user_id: user.id}}
```

No, that's absolutely not how I want to do it. That would be very, very hard to manage and hard to understand. Building up from smaller permissions means in this case that I can't create a good mental model of what the business rules look like.

There are some situations where it would make sense to build up from smaller permissions. This is not one of them. Please believe me when I say that I know that the most maintainable solution for this application would be the way I am currently doing it.

> You should redefine your sentence: `users can manage time entries on their own clients' projects, unless those time entries have already been invoiced` into
`users can update and destroy time entries of their own client's project if time entries are not already invoiced. users can always read time entries of their own client's project`

Your own example shows exactly why, in fact, I should *not* do this. Notice how much more complex your suggestion is than mine. Notice how much harder it is to understand. 

Why are you being so resistant to adding a new feature that will help make the `Ability` class easier to maintain for complex conditions?

Senjai:
>Why are you being so resistant to adding a new feature that will help make the Ability class easier to maintain for complex conditions?

We are performing our due diligence. This is a large decision as if it were to be implimented, it would have to be maintained. If the decision turned out to be poor for whatever reason, removing it is much more difficult then adding it. You don't need to bash here when we're just trying to understand if this is required.

--

In my opinion, this makes the ability class more complex. 

>I can manage a TimeEntry on my own projects unless the following conditions apply

Can be handled with:

```ruby
can :manage, TimeEntry, user_id: user.id
cannot :manage, TimeEntry, some_condition: :is_a_thing
```

I have not found any super complex situations where the above cannot be used effectively. Often, if rules are super complex it's a sign that refactoring could be done in your application to make things a little bit easier.

I still have outstanding comments about the consistency of the interface, and that not has to be at the top level and doesn't work across associations. That and being the first time the hash syntax doesnt have a one to one mapping with active records hash syntax for queries which adds complexity if we have to manipulate it before passing it to active record in the future, for whatever reason.

As @coorasse mentioned, I can't see a concrete example of why this would be necessary with the existence of cannot. Combined with an inconsistent interface I can't really support bringing this in. I've done large authorization audits of many projects without having to be concerned about incredibly complex rules. 

As an open source example: https://github.com/solidusio/solidus/pull/136

Thank you for your contribution and I appreciate your passion, but I'm going to have to shelve this until additional evidence, or perhaps a cleaner and more consistent interface is brought forward.

marnen:
@Senjai:
> We are performing our due diligence. 

I'm aware of that, obviously. I was simply asking what the argument against it was, because to me, the pull request seems to make the gem a lot easier to use and more capable for larger projects.

> If the decision turned out to be poor for whatever reason, removing it is much more difficult then adding it. 

That's a good point, but why do you think it might be a poor decision?

 > You don't need to bash here when we're just trying to understand if this is required.

No bashing was meant. I was simply asking a question.

> I can manage a TimeEntry on my own projects unless the following conditions apply
Can be handled with:
```ruby
can :manage, TimeEntry, user_id: user.id
cannot :manage, TimeEntry, some_condition: :is_a_thing
```

Yes, of course it can. But I think that that statement entirely misses the point of this pull request. I *like* the `can`...`cannot` form that you demonstrate here (@coorasse apparently doesn't, though). I am *using* that form in the [example I gave earlier](https://github.com/marnen/duckbill/blob/007c7ae4a26278222bbe3420b0c2557d24d6f3c7/app/models/ability.rb#L36-L37). This pull request is designed to make that form applicable in more cases.

This pull request, then, is designed to also allow me to do
```ruby
can :manage, TimeEntry, user_id: user.id
cannot :manage, TimeEntry, not: {some_condition: :is_a_thing}
```

> I have not found any super complex situations where the above cannot be used effectively. Often, if rules are super complex it's a sign that refactoring could be done in your application to make things a little bit easier.

That is a red herring. You seem to think I'm objecting to the `can`...`cannot` form, which suggests to me that you did not look carefully at the [example I supplied](https://github.com/marnen/duckbill/blob/007c7ae4a26278222bbe3420b0c2557d24d6f3c7/app/models/ability.rb#L36-L37). Please do so.

> I still have outstanding comments about the consistency of the interface, and that not has to be at the top level and doesn't work across associations. 

Then I will be happy to address those issues. What sort of interface would you like to see if this were to be incorporated?

(As for `not` not working across associations, I didn't need that feature, so I didn't implement it; consider it a first iteration.)

> As @coorasse mentioned, I can't see a concrete example of why this would be necessary with the existence of `cannot`. 

I gave such a [concrete example](https://github.com/marnen/duckbill/blob/007c7ae4a26278222bbe3420b0c2557d24d6f3c7/app/models/ability.rb#L36-L37). From your earlier comments, I believe you did not understand it, so I'll explain once more.

---
My example uses the `can`...`cannot` form you suggested above; that is, it starts with `can :manage, TimeEntry, project: {client: {user_id: user.id}}`, giving broad permissions for a class of objects.

Then, in the next line, it uses `cannot` to restrict this permissions for one special case, exactly as you suggested. It just happens that the condition on the `cannot` is negative, which is why I need the code from this pull request for this to work: `cannot [:destroy, :update], TimeEntry, not: {invoice_id: nil}`. I cannot replace `cannot not:` with `can`, because `can` never *restricts* permissions, but only broadens them.

In other words, you said above that you recommend the `can`...`cannot` form. That is what I am using in my example, and this pull request merely makes it usable in more cases.
---

> Combined with an inconsistent interface I can't really support bringing this in. 

Then I'll address the interface, or do whatever else it takes to get this into cancancan. It needs to be included if the gem is to work in a consistent, efficient fashion. Right now, the lack of negative conditions is a point of asymmetry that prevents cancancan from being as useful as it could be.

> I've done large authorization audits of many projects without having to be concerned about incredibly complex rules.
As an open source example: solidusio/solidus#136

I do not see what that has to do with the feature I am proposing.

jscheid:
Hi, I realize I'm a bit late to the party. We're in the same situation as @marnen in that we have a complex Ability file that would benefit tremendously from refactoring as described by them.

Basically we have a number of restrictions, for example that a data item must belong to the same customer account as the user, and need to combine them with more general access grants (e.g. users with role X can access data items of type Y).

Of course this can be done with vanilla CanCanCan, but it's complex, hard to maintain and relatively error prone.

It would be much clearer and safer if we could just add a rule at the bottom that said:

```ruby
cannot :manage, SomeModel, not: { account_id: user.account_id }
```

... rather than adding the opposite condition to each "positive" rule.

In terms of boolean logic, it would be the equivalent of turning `(A & !X) | (B & !X) | (C & !X)` into `(A | B | C) & !X`.

We'd like to stick to plain hash conditions (instead of scopes or blocks) for various reasons, including composability, performance and the ability to serialize the rules into JSON for use outside of our Rails app.

All this is just to say that this PR would definitely have been a valuable improvement from our POV 😄

Issue #102, opened 2014-07-24
Missing support for ActiveRecord::Enum (Rails 4.1)


tbk303:
Rails 4.1 introduced support for enums, represented by integers in the database. Currently, defining abilities matching enum values becomes tedious, as the actual enum value is required (for matching abilities on model instances) as well as the respective integer value (for matching abilities in database queries, i.e. accessible_by).

Example:

    can :read, FooBar, state: ['active', FooBar.states['active']]

Checking the ability against an instance with `authorize! :read, @foobar` requires the `'active'` value in the ability definition while fetching accessible foobars with `FooBar.all.accessible_by(current_ability)` requires the actual integer value (given via `FooBar.states`).

Maybe it is possible for CanCanCan to detect enum columns and extend the ability definition automatically.

markpmitchell:
@bryanrite, is there anything I can do to help with the process of reviewing my pull request for this issue?

bryanrite:
No.. sorry Mark, just been preoccupied with other things lately. Your PR looks good, I'll take a look at merging it tomorrow.

bryanrite:
Closing this in favour of #196

markpmitchell:
Thanks!

--
Mark Mitchell

markpmitchell:
Bryan —

It doesn’t look like you’ve had a chance to merge this PR yet. Is there anything I can do to help?

Thank you,

--
Mark Mitchell

bryanrite:
Sorry Mark, been having a hard time finding hours to continuing heading CCC maintainership. Looking for a new maintainer if you're interested :smile: 

Check out: #209

PR #196, opened 2015-04-01
Support ActiveRecord 4.x enums

markpmitchell:
Fixes Issue #102 

ActiveRecord queries for records with enums are by integer value, not by symbol. As CanCan uses the same query syntax, the way to specify a condition based on enum values is like so:

  can :read, Shape, color: Shape.colors[:blue]

While CanCan already generated the correct SQL queries for accessible_by, checks against individual objects did not work because the value of an enum attribute is a String. Therefore, the ActiveRecord 4 adapter must special case matches_condition? to convert the String to the appropriate integer before comparing against the value provided in the condition.

### DIFF
--- a/CHANGELOG.rdoc
+++ b/CHANGELOG.rdoc
@@ -1,5 +1,6 @@
 Develop
 
+* Fix cancancan#102 - Support for ActiveRecord 4.x enums (markpmitchell)
 
 1.10.1 (January 13th, 2015)
 
--- a/lib/cancan/model_adapters/active_record_4_adapter.rb
+++ b/lib/cancan/model_adapters/active_record_4_adapter.rb
@@ -18,6 +18,21 @@ module CanCan
         relation
       end
 
+      def self.override_condition_matching?(subject, name, value)
+        subject.class.defined_enums.include? name.to_s
+      end
+
+      def self.matches_condition?(subject, name, value)
+        # Get the mapping from enum strings to values.
+        enum = subject.class.send(name.to_s.pluralize)
+        # Get the value of the attribute as an integer.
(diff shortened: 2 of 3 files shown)


confiks:
I've been bumping into this issue as well. Thanks for the fix!

kimrgrey:
Hey, guys! What about this PR? I think it's pretty important patch because it's not possible to use cancancan with enums right now.

nelsonwittwer:
Would love to see more support for enums in CanCanCan. In the meantime, this is what I have been doing:

```ruby
can :read, ::Media, :published? => true
```

kimrgrey:
@nelsonwittwer, suppose that your approach will not work with `accessible_by` because cancan will try to make sql query with column `:published?` which not exists.

jwg2s:
:+1: would love to see this get merged

Senjai:
@jwg2s @markpmitchell I rebased your changes and pushed them to master. Thank you for your contribution! Sorry it took a while to get around to.

coorasse:
:+1:

kimrgrey:
Hey-ho! Great news! :heart:

markpmitchell:
Thanks!

--
Mark Mitchell

On Oct 7, 2015, at 8:19 PM, Richard Wilson <notifications@github.com> wrote:

@jwg2s <https://github.com/jwg2s> @markpmitchell
<https://github.com/markpmitchell> I rebased your changes and pushed them
to master. Thank you for your contribution! Sorry it took a while to get
around to.

—
Reply to this email directly or view it on GitHub
<https://github.com/CanCanCommunity/cancancan/pull/196#issuecomment-146406998>
.

PR #199, opened 2015-04-23
RubyDocs URL was directing to cancan which does not exist. Changed to cancancan

rpbaptist:
Just noticed the rdocs links went to a 404 page. Corrected the URL in the README.


bryanrite:
Thanks :)

Issue #98, opened 2014-07-20
Tackle the Beast (ControllerResource)


bryanrite:
Would love some help trying to tackle `ControllerResource` and breaking it down into smaller, SRP classes.

Let me know if you're interested!

gdpelican:
Hey @bryanrite, I might be interested in this, given a little guidance.

ssilvius:
what @bryanrite said.

PR #205, opened 2015-05-15
Controller resource refactor

gdpelican:
Hey @bryanrite so here's my stab at a refactor of ControllerResource for #98  which is, indeed, a beast. It's a modular redesign, with an aim to separate concerns as much as is possible in a place where the concerns are pretty tightly coupled (The method dependency tree for this class is a nest, heh.)

May need a couple tweaks here and there to get it to play nice with all the AR versions, but it's definitely at least ready for a look. Test suite runs green on both ends for me, and I plan on testing it out on a few moderately-sized projects in the next few days.


Senjai:
Hi @gdpelican, thank you for the effort here.

Unfortunately, this is not in a mergeable state for a couple reasons. Firstly, this is all done in one giant commit with no commit message. Refactors like this should be done iteratively, in very small commits with descriptive commit message detailing what you're doing and why you're doing it. This makes the history clean, allowing us to revert problem commits easily without reverting other parts of the code, or using git blame to determine why a certain decision was made.

All functionality you add should still have tests. It's great that this is passing the original test suite, but each module should have tests on its own. The goal is to simplify the controller resource test suite by encapsulating like behavior. 

While I'm happy to see that you're working on pulling behavior out into separate classes, there is a point where this concept can actually add complexity. For example, I cannot tell you what a Skipper does by reading the class name, and the class definition, without having to understand the other components in the system. In these cases, this can make code harder to change when you cannot easily identify the responsibilities of each component.

I would also recommend http://objology.blogspot.ca/2011/09/one-of-best-bits-of-programming-advice.html for some light reading. It's not a rule of thumb, but raises good points.

gdpelican:
Hooray a maintainer!

I'm happy to drive out some additional unit tests (there's no new functionality here, but happy to add some more coverage around the individual modules) and make the class names more expressive, but the (many) commits and messages have unfortunately been lost to squashing. I don't necessarily agree with the idea that a long, detailed history filled with developer hiccups, backtracks, experiments, etc. is a cleaner, better history, but maybe an update to CONTRIBUTING.md would prevent it from happening to someone else down the line? (Some projects enforce a pretty strict squashing policy, while others are more lenient.)

I guess my real question is whether you'd be interested in supporting this change going in (given appropriate names, test coverage, etc.), or whether it'd be a waste of time to develop further? I understand it's a big-ish change, but I think it's a pretty large step forward for developing on the gem going forward, and it's been noted in [several](https://github.com/CanCanCommunity/cancancan/issues/98) [places](https://github.com/CanCanCommunity/cancancan/pull/186) that this is a monolithic class that needs some serious breakdown. (Check out the codeclimate, if you're into that sort of thing:
cancanrefactor

freemanoid:
@gdpelican thank you for such a big effort.

As for me it's really hard to review such a big diff. You don't need to add the commits with all your "hiccups, backtracks, experiments, etc." because it's useless. But it would be much easier for others to review the changes in small commits with descriptive commit messages. You just need to separate all your changes into several logical parts.

BTW some of your files don't have a new line at the end.

gdpelican:
Hey @Senjai @freemanoid I've renamed some of the modules, slowly added some unit test coverage and doing some additional streamlining on this PR over the past while; should be ready for another look. Happy to hear any feedback, or any questions about the implementation.

gdpelican:
Such a shame no one's willing to take the time to take a look at this, it's really a nice improvement; I've been using it in my personal projects and haven't had any problems. :(

Senjai:
Thanks for the attempt here, I don't think this is the direction we want to go.

Issue #200, opened 2015-04-25
"can :create, Model do" broken?


svoop:
Given the following ability:

    can :create, Page do
      user.has_role?(:manager)
    end

The following has worked in the past:

    u = User.first
    u.has_role?(:manager)     # => false
    a = Ability.new u
    a.can? :create, Page   # => false

However, I just figured out that this no longer holds. The block passed to `can` in the definition is not hit unless you pass an instance, so this is what it looks like now:

    a.can? :create, Page       # => true - OOPS!
    a.can? :create, Page.new   # => false - OK

Is this a bug or a feature?

eikes:
I just noticed this as well. It believe, that this is due to the ```!subject_class?(subject)``` condition of the first elsif clause [line 33 in rule.rb](https://github.com/CanCanCommunity/cancancan/blob/5fa747dd5441701b48356647d53ca213f7ca2e21/lib/cancan/rule.rb#L33) 

```ruby
    # Matches the block or conditions hash
    def matches_conditions?(action, subject, extra_args)
      if @match_all
        call_block_with_all(action, subject, extra_args)
      elsif @block && !subject_class?(subject)
        @block.call(subject, *extra_args)
      elsif @conditions.kind_of?(Hash) && subject.class == Hash
        nested_subject_matches_conditions?(subject)
      elsif @conditions.kind_of?(Hash) && !subject_class?(subject)
        matches_conditions_hash?(subject)
      else
        # Don't stop at "cannot" definitions when there are conditions.
        conditions_empty? ? true : @base_behavior
      end
    end
```

The example given in the [Ability.rb](https://github.com/CanCanCommunity/cancancan/blob/5fa747dd5441701b48356647d53ca213f7ca2e21/lib/cancan/ability.rb#L39-L46) won't work under these circumstances either:

```ruby
    # Any additional arguments will be passed into the "can" block definition. This
    # can be used to pass more information about the user's request for example.
    
    can? :create, Project, request.remote_ip
    
    can :create, Project do |project, remote_ip|
      # ...
    end
```

eikes:
I wrote a test that verifies that this does not work. Pull request is here: https://github.com/CanCanCommunity/cancancan/pull/206

Senjai:
@svoop @eikes I've been able to reproduce this easily. Thanks for your example eikes.

I noted there are specific tests guarding against this behavior, @svoop do you remember when this has worked for you or what version this worked in?

svoop:
[...]

PR #206, opened 2015-05-19
Create ability test on a Class subject with a block which returns false

eikes:


eikes:
This test fails even though it shouldn't. The related issue is here: https://github.com/CanCanCommunity/cancancan/issues/200

Senjai:
Closing per conversation on #200

Issue #169, opened 2015-01-17
List current_user abilities


devaroop:
A list of current_user's abilities would really help client side apps(like AngularJS, EmberJS) to paint the UI based on that. The list could include default CRUD operations along with custom options from the ability class.

Please let know if this is desired, I will be happy to develop and raise a pull-request.

Abdenasser:
+1

Stormcore:
Awesome! Really need this feature!

caseywatts:
This sounds really useful!

giedriusr:
@devaroop I just wanted to create a new issue/feature regarding this. This is really useful in case you described above. Will you start working on this?

devaroop:
Sure, will be happy to contribute. Starting...

giedriusr:
:+1: awesome

Senjai:
Closing to focus additional conversation on #208

Issue #195, opened 2015-03-31
List of things user can?


firedev:
Hi,

This is probably related to #169. How one would list everything user `can? :manage` for example?

Here is a concern for `Ability` I came up with, but I am sure there must be a better way:

```ruby
# models/concerns/can_tell_what_it_can_do
module CanTellWhatItCanDo
  extend ActiveSupport::Concern

  def can_do(action)
    rules.map do |r|
      next unless r.actions.include? action
      r.subjects.flat_map(&:name)
    end.flatten.compact.uniq
  end
end

current_ability.can_do(:manage)
=> ["Page", "User"]
```

nicolas-besnard:
Is `current_ability` the response ?

firedev:
Thanks. Anything I can improve apart from that?

alperkokmen:
hi @firedev, have you made any progress with your implementation? i am looking for something similar to use on client-side?

firedev:
Nope, that's what I am using at the moment.

ogerman:
Here is small code snippet, which implements abilities map.
https://github.com/runemadsen/cancan-backbone/blob/master/README.md#setup

Senjai:
Closing to focus discussion on #208

PR #208, opened 2015-05-26
[Solved] list user abilities

devaroop:
Resolves issues #169 and #195 

Three methods created for ease:

1) list_all
2) list_can(filter_action = nil)
3) list_cannot(filter_action = nil)

Will add docs once pull request is approved. Thanks.

### DIFF
--- a/lib/cancan/ability.rb
+++ b/lib/cancan/ability.rb
@@ -356,7 +356,7 @@ module CanCan
     end
 
     def list_abilities
-      abilities_list = {can: {}, cannot: {}}
+      abilities_list = {:can => {}, :cannot => {}}
 
       rules.each do |rule|
         subjects = rule.subjects
--- a/lib/cancan/ability.rb
+++ b/lib/cancan/ability.rb
@@ -252,18 +252,23 @@ module CanCan
       self
     end
 
-    def list_can(filter_action = nil)
-      can_do_list = list_abilities[:can]
-      filter_action ? can_do_list[filter_action] : can_do_list
-    end
+    def permissions
+      permissions_list = {:can => {}, :cannot => {}}
 
-    def list_cannot(filter_action = nil)
-      cannot_do_list = list_abilities[:cannot]
-      filter_action ? cannot_do_list[filter_action] : cannot_do_list
-    end
+      rules.each do |rule|
+        subjects = rule.subjects
+        expand_actions(rule.actions).each do |action|
+          if(rule.base_behavior)
+            permissions_list[:can][action] ||= []
+            permissions_list[:can][action] += subjects.map(&:to_s)
+          else
+            permissions_list[:cannot][action] ||= []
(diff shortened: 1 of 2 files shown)


lib/cancan/ability.rb
```diff
@@ -252,6 +252,20 @@ def merge(ability)
       self
     end
 
+    def list_can(filter_action = nil)
```
Senjai: I'm not sure we need list_can, or a filter_action. As well as the same for list_cannot. It just wraps a hash we have to build in memory anyway.

lib/cancan/ability.rb
```diff
@@ -252,6 +252,20 @@ def merge(ability)
...
+      can_do_list = list_abilities[:can]
+      filter_action ? can_do_list[filter_action] : can_do_list
+    end
+
+    def list_cannot(filter_action = nil)
+      cannot_do_list = list_abilities[:cannot]
+      filter_action ? cannot_do_list[filter_action] : cannot_do_list
+    end
+
+    def list_all
```
Senjai: list_all just aliases to a private method. Instead we should just have one public method.

lib/cancan/ability.rb
```diff
@@ -340,5 +354,25 @@ def default_alias_actions
         :update => [:edit],
       }
     end
+
+    def list_abilities
```
Senjai: I'd be :+1: in making this public and making naming it to something like `permissions` or something a little more indicative that it provides a hash of permissions rather than a list of `Ability`s to avoid confusion.

lib/cancan/ability.rb
```diff
@@ -340,5 +354,25 @@ def default_alias_actions
...
+
+      rules.each do |rule|
+        subjects = rule.subjects
+        expand_actions(rule.actions).each do |action|
+          if(rule.base_behavior)
+            abilities_list[:can][action] ||= []
+            abilities_list[:can][action] += subjects.map {|sub| sub.respond_to?(:name) ? sub.name : sub}
+          else
+            abilities_list[:cannot][action] ||= []
+            abilities_list[:cannot][action] += subjects.map {|sub| sub.respond_to?(:name) ? sub.name : sub}
```
Senjai: Do we care if it responds to name? Can we simply call to_s on whatever it is?

Senjai:
I'm :+1: on this. It seems to be a desired feature and has little to no risk. It will require a changelog entry. 

Can you address the feedback I've left as well as squash this down into a single commit instead of two?

devaroop:
Sure, will have them ready in a couple of days. Thanks.

giedriusr:
:+1:

devaroop:
@Senjai @giedriusr Guys there were problems with squashing because there were latest updates which crept in. I opened a new pull request #221 which is ready to merge. Closing this one now.

giedriusr:
:+1:

nazartaran:
Hey, everyone, when that stuff will be merged ?

PR #225, opened 2015-06-29
Fix #214 "Cannot" with association hash conditions : accessible_by generates incorrect SQL (Postgres / ActiveRecord)

### LINKED ISSUE #214: "Cannot" with association hash conditions : accessible_by generates incorrect SQL (Postgres / ActiveRecord)
First of all, thanks for this gem of a gem ;)

I just stumbled into something that I think might be an edge case, resulting in a bug (at least on postgresql).

I have the following setup (a bit modified for clarity) : 

``` ruby
class Organization < ActiveRecord::Base

   belongs_to :owner, class_name: :Account
   has_many :staff_memberships

end

class StaffMembership < ActiveRecord::Base

  belongs_to :organization
  belongs_to :account

end

class Ability
  include CanCan::Ability

  def initialize(user)
    can :manage, Organization, owner_id: user.id
    can :manage, Organization, staff_memberships: {account_id: user.id}
    # here is the offending line
    cannot :discuss_as, Organization, staff_memberships: {account_id: user.id}
  end

end
```

When I run `Organization.accessible_by(Ability.new(Account.find 1))`, the following SQL is generated (modified for clarity) : 

``` sql

SELECT organizations.*
FROM    organizations
LEFT OUTER JOIN staff_memberships
ON staff_memberships.organization_id = organizations.id
WHERE not (staff_memberships.account_id = 1)
AND ((staff_memberships.account_id = 1) OR (organizations.owner_id=1));

```

The problem here lies with `not (staff_memberships.account_id = 1)`.

If no staff_membership is joined to the record, `staff_memberships.account_id` is `NULL`, so the conditions evaluates to `NULL` and the record is filtered out, which it shouldn't.

Moreover, this SQL would incorrectly select an organization where a staff_membership with account_id 1 exists, if other staff_memberships exist (because of the left outer join)

The correct SQL would use an anti-join, for example : 

``` sql
WHERE NOT EXISTS (
  SELECT FROM staff_memberships 
  WHERE account_id = 1 
  AND organization_id = ?
)
```

**Is there something i'm doing wrong, or is it a feature that is not supported by CCC ? Do I have to use a scope condition in this case ?**

note: I tracked the origin of the offending SQL down to this line : https://github.com/CanCanCommunity/cancancan/blob/develop/lib/cancan/model_adapters/active_record_adapter.rb#L106

note : for the record, here is a workaround using SQL fragments: 

``` ruby
# This is the relevant SQL fragment, generated using AREL
has_membership = StaffMembership.where(account: user).where(organization_id: Organization.arel_table[:id]).exists.to_sql
# matching block
block = ->(organization){ organization.staff_memberships.any?{|m| m.account_id == user_id }}

cannot :discuss_as, Organization, has_membership, &block
```

tiokksar:
Fixed issue of NULL values being filtered out by cannot definitions.

### DIFF
--- a/lib/cancan/model_adapters/active_record_adapter.rb
+++ b/lib/cancan/model_adapters/active_record_adapter.rb
@@ -14,7 +14,7 @@ module CanCan
       #   can :manage, User, :id => 1
       #   can :manage, User, :manager_id => 1
       #   cannot :manage, User, :self_managed => true
-      #   query(:manage, User).conditions # => "not (self_managed = 't') AND ((manager_id = 1) OR (id = 1))"
+      #   query(:manage, User).conditions # => "(not (self_managed = 't') OR (self_managed = 't') IS NULL) AND ((manager_id = 1) OR (id = 1))"
       #
       def conditions
         if @rules.size == 1 && @rules.first.base_behavior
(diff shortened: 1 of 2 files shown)


coorasse:
can you please give me a practical example of when this happens? I never saw using IS NULL this way.

coorasse:
I close due to lack of feedback. Feel free to reopen with a detailed explanation of the error

PR #233, opened 2015-07-14
Correct information about strong parameters.

riffraff:
Update README based on the fact that CanCan doesn't actually
do anything special with `update_params`, as per discussion in #184.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,9 +89,26 @@ See [Authorizing Controller Actions](https://github.com/CanCanCommunity/cancanca
 
 When using `strong_parameters` or Rails 4+, you have to sanitize inputs before saving the record, in actions such as `:create` and `:update`.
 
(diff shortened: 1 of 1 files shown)


Senjai:
@riffraff Thank you, this should clear up some confusion

graywh:
2.5 years later and the wiki still hasn't been updated with this change

newfylox:
I still don't understand why we would need an `update_params` method if we need to call it manually. What's the difference if I'm using another name like `ham_params` when doing `@instance.update(ham_params)` ?

Also yeah, in the doc, we can still read this:

https://github.com/CanCanCommunity/cancancan/blob/1100093bd2d3416bb875e1cb407a81ec35a6dab5/docs/controller_helpers.md?plain=1#L170

PR #234, opened 2015-07-27
Update travis to use container-based infrastructure

dgynn:
The Travis build config can be updated to allow the tests to run on their container-based infrastructure.
http://docs.travis-ci.com/user/migrating-from-legacy/

This lets the tests start faster and run quicker. This PR should be compatible with #230 which would also be great to get merged.


Senjai:
@dgynn Thanks!

Issue #228, opened 2015-07-06
Eager loading on member resource


lowjoel:
Hi there!

I'm using the [Bullet](https://github.com/flyerhzm/bullet) gem which warns when my resources omit includes.

For collection resources, the collections are lazy-loaded, so I can still call `@collection = @collection.includes(:relation)` in any given action.

For member resources, the resource is already loaded by the time is action is called. Is there any way to specify that the resource should have certain things eager loaded without resorting to loading it manually?

Maybe, `load_resource :resource_name, eager_load: :relation`... and it would be loaded for both collection and member resources? It would appear that modifying `ControllerResource#resource_base` would be sufficient to implement this.

I find duplicating Cancancan's logic for determining whether it is a collection or member route, then remembering the resources to load through, isn't particularly DRY.

Comments/ideas on how to move ahead?

Senjai:
Closing in favor of all the other eager loading related PR's open.

PR #238, opened 2015-08-10
Feature/remove eager loading

coorasse:
Using code from @elabs and continuing on what I wrote here: https://github.com/CanCanCommunity/cancancan/pull/171#issuecomment-129334100 that is my suggestion.
Note that I fixed it only for AR4


lib/cancan/model_adapters/active_record_adapter.rb
```diff
@@ -65,6 +65,7 @@ def database_records
           @model_class.where(nil).merge(override_scope)
         elsif @model_class.respond_to?(:where) && @model_class.respond_to?(:joins)
           if mergeable_conditions?
+            puts conditions
```
bmulholland: You'll want to remove the outputs in this file :)

coorasse: :+1:

Senjai:
Could we write some more tests to verify the correct sql is being generated? This is a more complex PR and may have some edge cases I may not be aware of. Going to pull it down and review it more thoroughly.

coorasse:
Would you prefer to reduce the amount of code and depend directly from this gem? https://github.com/elabs/ar_outer_joins

Is an old gem with just 15 stars so I preferred to include the code directly but we can eventually remove it and add the dependency. what do you think?

mrloop:
**Configurable includes** In some scenarios I'd like to be able to specify per model what gets included to reduce further sql requests I know I'm going to make when looking at a resources related models.

mrloop:
Ah look somebody has already suggested this #228

coorasse:
You can specify what to include already with `includes(:table_a, :table_b)` directive. Which is even more clear.

mrloop:
@coorasse but I'm loading the resource via `load_and_authorize_resource`. How do I tell cancan to `.includes(:table_a, :table_b)` during `load_and_authorize_resource` find?

mrloop:
Ok, your just getting a scope from load_and_authorize_resource, thanks @coorasse

coorasse:
Hello @Senjai. I think this PR brings an important improvement and, at the same time, change in cancancan.
I can rely on this gem https://github.com/elabs/ar_outer_joins in order to avoid writing more tests here.

I already use the following changes in different projects and I had no problems, even in complex ones (https://github.com/coorasse/Airesis/tree/master/app/cancan).

I really think using `includes` to perform outer_joins is a bad choice and cancancan is taking the responsability of preloading stuff when is not required --> performance problems in case of deep relations.
Let the user preload their stuff and let cancancan just check for permissions.

derosm2:
+1

lowjoel:
@coorasse https://github.com/nerde/left_join seems a little newer. Do you want to use that instead?

Senjai:
@coorasse Want to get on a hangout at some point for this one?

coorasse:
Hi @Senjai, get in touch with me via hangout at coorasse@gmail.com if you want

kaspernj:
+1

kaspernj:
Once the following PR is merged into Rails, it should be possible to just do a `.left_outer_joins` instead in Rails 5: https://github.com/rails/rails/pull/30410

aliou:
Any news on this feature ?

coorasse:
I think this will fall in the next Major version (3.0) where I plan to deprecate also AR4. Therefore this PR will be simplified by a lot and rely on left_outer_join by AR5

lib/cancan/model_adapters/ar_outer_joins/ar_outer_joins.rb
```diff
@@ -0,0 +1,15 @@
+require "cancan/model_adapters/ar_outer_joins/join_builder"
+require "cancan/model_adapters/ar_outer_joins/join"
+
+module ArOuterJoins
+  def outer_joins(*joins)
```
kaspernj: In Rails 5 we got `.left_joins`. The duplicate table in this regard was fixed in https://github.com/rails/rails/pull/30410.

Maybe rename this to `left_joins` and only use if it Rails version is below Rails 5.2?

bmulholland:
What's the status of v3.0?

coorasse:
Closed in favor of #508

PR #241, opened 2015-08-18
fixing example of authorize of README

TheBlasfem:
i think that in the example of `authorize!`, the first variable should have the same name of the method where is executed. It keeps more easy to understand `authorize! :show, @article` instead of `authorize! :read, @article`


TheBlasfem:
i'm going to close this, i realized that my comment was wrong

PR #242, opened 2015-08-18
Allow Symbol or Proc to define model class

inkstak:
I have a particular case, where the `:class` option for `load_and_authorize_resource` depends on various conditions, evaluated in the context of the controller instance.

I suggest to support symbol and/or proc for setting this options : 

```
class FooController < ApplicationController
  load_and_authorize_resource :foo, class: :model

  def model
    whatever? ? Foo1 : Foo2
  end
end
```

```
class FooController < ApplicationController
  load_and_authorize_resource :foo, class: proc { whatever? ? Foo1 : Foo2 }
end
```

What do you think ?

NOTE: This PR is a duplicate of #239, sorry!

### DIFF
--- a/lib/cancan/controller_resource.rb
+++ b/lib/cancan/controller_resource.rb
@@ -149,6 +149,8 @@ module CanCan
       when false  then name.to_sym
       when nil    then namespaced_name.to_s.camelize.constantize
       when String then @options[:class].constantize
+      when Symbol then @controller.send(@options[:class])
+      when Proc   then @options[:class].call(@controller)
       else @options[:class]
       end
     end
--- a/spec/cancan/controller_resource_spec.rb
+++ b/spec/cancan/controller_resource_spec.rb
@@ -441,6 +441,25 @@ describe CanCan::ControllerResource do
       expect(controller.instance_variable_get(:@model)).to eq(model)
     end
 
+    it "loads the model using a symbol" do
+      model = Model.new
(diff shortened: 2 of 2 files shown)


Senjai:
I'm not particularly against the code, but I'm curious as to why you wouldn't know the class at runtime? In what scenario does this need to occur? I would think that there might be a better solution to whatever it is you're trying to solve instead of having a controller dynamically handle numerous resources..

inkstak:
I've a legacy database, with numerous key/value models (>50).
Admins need to be able to create/update/delete them.

Instead of creating a controller for each model, I've factored one controller:

    class LabelsController < ApplicationController
      load_and_authorize_resource :label, class: :resource_model, parent: false

      ..

      protected

      def resource_model
        "Label::#{ params[:model] }".constantize
      end
    end

Senjai:
Right, I would consider than an exceptional case. If a model just has a key, and a value, it shouldn't be a model. I would probably suggest finding a method of reducing the number of models in your system to fix the problem rather than attempt to fix it here. Also, your code is vulnerable to remote code execution as an FYI.

inkstak:
Thanks for your advices, but refactoring these models is not possible at this state (data used by third services) despite the fact that I'll love to have only one model. Btw, my code is more robust than the provided example.

At the end, I'll use a `before_action` method in place of  `load_and_authorize_resource`.
Thanks for your time.

Issue #255, opened 2015-10-08
Possible regression with draper when updating from 1.12


dgilperez:
Hi!

I'm using devise + cancancan along with draper gem, and until 1.12 the behavior was that calling `can?` on a decorated model used the decorated model for checking authorization instead of the decorator class. Maybe I can better explain myself with an example: 

     # in my ability.rb
     can :read, User do |user|
       true
     end

     # in a view
     @user = User.last.decorate # instance of UserDecorator
     can? :show, @user 
     #=> true, since the ability for User was used
      
After updating to 1.13, this is no longer the case, and calling `can? :show, @user` on a decorated user will not use `User` ability at all.

Is this intended behavior or is it a regression? If this is intended, how can I get that bahaviour back, maybe with an extension or configuration in my decorators / ability.rb?

Thanks!

Senjai:
@dgilperez Definitely not intended. Could you reproduce this using a `SimpleDelegator` from the ruby standard library? That should behave similarly to draper, if not it might be draper specific. It would be helpful if we had a regression for this with a delegator so this wouldn't happen in the future.

Senjai:
I would expect this might be related to https://github.com/drapergem/draper/blob/master/lib/draper/decorator.rb#L191-L204, not being checked with the new performance rule indexing stuff.

dgilperez:
Yeah, let me play around and see if I can add a failing test to the suite.

dgilperez:
BTW, yeah, that'd be the workaround, though that's really inconvenient in a large app (like mine :().

dgilperez:
Test added (or so I think), please let me know.

jaredbeck:
I believe I've run into the same issue, using draper 2.1.0, cancancan 1.13.1, rails 4.2.4.

```
@redacted #=> #<RedactedDecorator:0x00..a0 @object=#<Redacted:0x00..18>>
can?(:update, @redacted) #=> false
can?(:update, @redacted.object) #=> true
```

Downgrading to cancancan 1.12.0 fixes the issue.

Senjai:
@jaredbeck @dgilperez I don't think this problem falls within CanCanCan's domain. The performance fix works with the ancestors of an object, which is a valid thing to check when asserting that something is something else.

If something like Draper is merely pretending to be something else, but doesn't satisfy this requirement completely, it isn't fair to add additional logic at the cost of performance to ensure we cover that edge case.  

[...]

PR #257, opened 2015-10-08
Handle delegated classes - fixing regression with Drapper

dgilperez:
This is WIP, addressing #255 

So this would be the failing test. How do you suggest moving forward here, @Senjai?

### DIFF
--- a/spec/cancan/ability_spec.rb
+++ b/spec/cancan/ability_spec.rb
@@ -361,6 +361,20 @@ describe CanCan::Ability do
     expect(@ability.can?(:read, A)).to be(true)
   end
 
+  it "allows to check ability in a decorated class" do
+    class ADecorator < SimpleDelegator
+      def kind_of?(klass)
+        __getobj__.kind_of?(klass)
+      end
+      alias_method :is_a?, :kind_of?
+    end
+    class A; end
+
+    @ability.can :read, A
+    expect(@ability.can?(:read, A.new)).to be(true)
+    expect(@ability.can?(:read, ADecorator.new(A.new))).to be(true)
+  end
+
   it "checks permissions through association when passing a hash of subjects" do
     @ability.can :read, Range, :string => {:length => 3}
 


Senjai:
@dgilperez Does this pass in 1.12?

spec/cancan/ability_spec.rb
```diff
@@ -361,6 +361,20 @@ class A; include B; end
     expect(@ability.can?(:read, A)).to be(true)
   end
 
+  it "allows to check ability in a decorated class" do
+    class ADecorator < SimpleDelegator
+      def kind_of?(klass)
```
Senjai: We shouldn't have to override kind_of. I'm not sure we want to support things that override kind_of. It's inflexible. I can look at what's required, but draper should not be overriding this, as automatic delegation is an option not a default.

dgilperez: Yeah, I agree with your view, but that's how actually draper works regarding this case, isn't it? I just replicated was Draper was doing ... since SimpleDelegator and Draper work in different ways, what should we do here? Check for `defined?(:__get_object__)` or `.object`? Not cool either ... kind of lost here.

robmathews: I wonder if we can poke at Draper, change the way it is doing this instead. We could push a pull request over there just as easily.

robmathews:
I think we're all using Draper the wrong way. From the documenation below, I think we should just be using 

   decorates_assigned :article

See the documentation here:

Because decorators are designed to be consumed by the view, you should only be accessing them there. Manipulate your models to get things ready, then decorate at the last minute, right before you render the view. This avoids many of the common pitfalls that arise from attempting to modify decorators (in particular, collection decorators) after creating them.

    To help you make your decorators read-only, we have the decorates_assigned method in your controller. It adds a helper method that returns the decorated version of an instance variable:

    # app/controllers/articles_controller.rb
    class ArticlesController < ApplicationController
      decorates_assigned :article

      def show
        @article = Article.find(params[:id])
      end
    end
    The decorates_assigned :article bit is roughly equivalent to

    def article
      @decorated_article ||= @article.decorate
    end
    helper_method :article

Senjai:
@robmathews I agree that `@article` should indeed be an Article. But this doesn't solve the problem people are having if you use `can? :do_things, @decorated_article` Which is the behaviour they want to be able to support.

robmathews:
@Senjai - ya, I agree. My advice only really helps people going forward, and I quickly realized it doesn't much help me.

dgilperez:
@robmathews +1 on that, thanks in deed for pointing that out. I never really had a strong opinion about using `decorates_assigned` instead of `@article = @article.decorate` ... now I get it. Anyway, no way back for me at the moment, it'd be too much cost for the gain.

robmathews:
I hear you - sort of struggling with the same issue myself.

robmathews:
So the DraperDecorator solution is pretty much useless, since you might well decorate an object and then use CanCan in a view like this:

    = link_to_if can?(:destroy, foo), “Delete foo”, foo_path

Now you are in a view, so it's ok to decorate (according to Draper) but  it won't work since you are trying to use CanCan. 

Seems like the official Draper solution is pretty much non-functional.

robmathews:
I'm working on a gem that connects cancancan and draper in a reasonable way. I'll post the reference here when I finish. Shouldn't be long.

Senjai:
Closing to limit conversation to #255

PR #258, opened 2015-10-08
Fix #merge with empty Ability

jhawthorn:
Using `merge` with an empty ability caused an error due to `rules.each` being `nil.each`.

This moves the default value of `@rules` to the `#rules` method.


Senjai:
:+1:

PR #260, opened 2015-10-14
Point to StackOverflow

coorasse:
Point questions to stack overflow


Senjai:
@coorasse Thanks! This should help.

PR #265, opened 2015-10-24
Execute SQL when evaluating a conditional rule on an AR scope

lowjoel:
When we are evaluating arguments within the context of an AR scope, execute the SQL instead of loading all rows to memory.

Typically we would be using this on records where the association is not loaded to memory because there are too many. Cancancan will currently load all the association's rows then evaluate against the conditions.

If the old behaviour of checking conditions directly is desired, then the collection can be eager loaded instead. Loaded collections will not execute another SQL query.

This only affects AR4, since AR3 does not allow us to obtain an association proxy.

Comments welcome.

### DIFF
--- a/lib/cancan/model_adapters/active_record_4_adapter.rb
+++ b/lib/cancan/model_adapters/active_record_4_adapter.rb
@@ -6,6 +6,37 @@ module CanCan
         model_class <= ActiveRecord::Base
       end
 
+      def self.override_association_matching?(subject, conditions)
+        subject.kind_of?(ActiveRecord::Relation) && !subject.loaded?
+      end
+
+      def self.matches_association?(subject, conditions)
+        where_conditions = association_conditions_for(subject.proxy_association.reflection, conditions)
+        subject.where(where_conditions).any?
+      end
+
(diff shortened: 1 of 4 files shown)


lib/cancan/model_adapters/active_record_4_adapter.rb
```diff
@@ -6,6 +6,37 @@ def self.for_class?(model_class)
...
+          conditions.each_pair do |key, value|
+            key, value = association_condition_for(target, key, value)
+            result[key] = value
+          end
+        end
+      end
+
+      def self.association_condition_for(target, key, value)
+        string_key = key.to_s
+        if ActiveRecord::VERSION::MINOR >= 1 && target.defined_enums.include?(string_key)
```
lowjoel: @Senjai Potentially in a future version maybe we should have one class each for AR4/AR4.1/AR4.2/AR5? AR5 won't need the enum hack, for one...

lowjoel:
Hi @Senjai did you have a chance to have a look at this?

lowjoel:
This is implemented in my Squeel adapter, see #307.

PR #266, opened 2015-10-29
Optimize performance of `cancan_resource_class`

kirs:
No need to stringify all ancestors if InheritedResources is not even used in the project.

### DIFF
--- a/lib/cancan/controller_additions.rb
+++ b/lib/cancan/controller_additions.rb
@@ -280,7 +280,7 @@ module CanCan
       end
 
       def cancan_resource_class
-        if ancestors.map(&:to_s).include? "InheritedResources::Actions"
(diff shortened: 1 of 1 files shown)


coorasse:
What about getting rid of it?

kirs:
@coorasse what do you mean?

coorasse:
I was just wondering if we should get rid the support to InheritedResource since is not maintained anymore. I don't like to keep supporting libraries which are old and not maintained. But your PR is fine :+1:

kirs:
Sounds like a good idea.

waiting-for-dev:
Any news on that? I have been doing some [benchmarks with derailed_benchmarks and stackprof](https://github.com/schneems/derailed_benchmarks#my-app-is-slow) in one of our apps and, in some cases, this method accumulates 15% of CPU pct.

Thanks

coorasse:
Closed in favour of #370

PR #270, opened 2015-11-03
Fix build_resource when model name is Action

### LINKED ISSUE #125: When controller occurs before application controller
I have the issue: 

```
undefined method `each_pair' for "new":String
# lib/mongoid/attributes/processing.rb:23:in `process_attributes'
```

The issue disappears when: 
- Removing `load_and_authorize_resource` and `check_authorization` from the controller (with the latest versions of devise(3.3.0) & cancancan(1.9.2) )
- When the controller is renamed so that it occurs alphabetically after the `application_controller.rb`.

To see this error run this app: https://github.com/mrjlynch/sample-mongoid-error-app and go to `localhost:3000/action/new`

anilmaurya:
- Fix #125

### DIFF
--- a/lib/cancan/controller_resource.rb
+++ b/lib/cancan/controller_resource.rb
@@ -240,7 +240,8 @@ module CanCan
       elsif @options[:class] && @params.has_key?(extract_key(@options[:class]))
         @params[extract_key(@options[:class])]
       else
-        @params[extract_key(namespaced_name)]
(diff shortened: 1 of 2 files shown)


alexsanford:
This PR also fixes the issue for me, although I get a different error message than the one in the original issue. The error message I get is `When assigning attributes, you must pass a hash as an argument.`. In any case, this PR fixes it 😄

PR #271, opened 2015-11-04
Fix attributes_for result for hash conditions

### LINKED ISSUE #108: attributes_for not giving the full hash of conditions
Hi,

I've been playing around with `attributes_for` and it seems this doesn't give back the real hash of conditions one can expect.

For instance, I've defined a rule like:

```
can :manage, [OwnedObject], { owner: { parent_id: @user.id } }
```

I've been the calling 

```
@user.ability.attributes_for(:index, OwnedObject)
```

And i get back an empty Hash.
However, if i define the following rule: 

```
can :manage, [OwnedObject], { owner_id: @user.id }
```

I will get the expected hash.

I think this is due to `rule.attributes_from_conditions` which is called by `attributes_for` which adds the rule `unless [Array, Range, Hash].include? value.class`.

FYI, the reason i would like to use `attributes_for` is that I'm using the elasticsearch-rails gem as an alternate source of data and i'd like to apply to my searches the same conditions as my models' queries. For that i need the hash of conditions declared in my ability class.

Thanks

anilmaurya:
fix #108

### DIFF
--- a/lib/cancan/rule.rb
+++ b/lib/cancan/rule.rb
@@ -70,7 +70,7 @@ module CanCan
     def attributes_from_conditions
       attributes = {}
       @conditions.each do |key, value|
-        attributes[key] = value unless [Array, Range, Hash].include? value.class
+        attributes[key] = value unless [Array, Range].include? value.class
       end if @conditions.kind_of? Hash
       attributes
     end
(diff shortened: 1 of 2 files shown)


coorasse:
I do not see any wide need for this. Thanks for your effort

PR #274, opened 2015-11-14
Make the following fixes

### LINKED ISSUE #177: Rule#matches_conditions_hash: return used in enumerable.all? block
In `Rule#matches_conditions_hash` [there's](https://github.com/CanCanCommunity/cancancan/blob/75b9de36796ba0291710f33295780e1f15aa4ac8/lib/cancan/rule.rb#L111) a `return` statement in a block for `Enumerable#all?`, which causes it to not check other conditions (introduced by 7a60774a5076ec09f7185d3cff217ac0557544f8). Did i miss something? I guess the fix is just to replace it with an if-else.

abhishekbose87:
- Fixes #177 
- Fix spec/README.rdoc for running individual specs

### DIFF
--- a/lib/cancan/rule.rb
+++ b/lib/cancan/rule.rb
@@ -112,10 +112,10 @@ module CanCan
 
       conditions.all? do |name, value|
         if adapter.override_condition_matching?(subject, name, value)
-          return adapter.matches_condition?(subject, name, value)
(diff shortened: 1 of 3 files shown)


Senjai:
@abhishekbose87 Thank you

abhishekbose87:
@Senjai :)

Issue #146, opened 2014-11-25
load_and_authorize_resource fails with 2 level modules


darthjee:
when running `load_and_authorize_resource` on a 2 module level nestled controller, it raises a `NameError`

``` ruby
class BackOffice::RenewalBulk::ImportsController
  load_and_authorize_resource
  ...
end
```

> Failure/Error: send(http_verb, action, try(:params))
>      NameError:
>        uninitialized constant Import

It has been found while running

```
'rails', '~> 4.1.7'
'cancancan', '~> 1.9.2'
```

PR #275, opened 2015-11-18
Fix attempt to #constantize a symbolic name

bambycha:
Related with #146, 6c497b8.

I run into this problem when implementing a successive version of a resource.
I was using something like next in controller:

```ruby
class V2::InvoiceController < V2::BaseController
  load_and_authorize_resource :invoice
end
```

Pretty similar to what docs suggests. But regardless of existing logic that try to compose namespaced resource. It fails due to passing a name of the resource as `:invoice` symbols. And rescuing with a not namespaced resource.

```ruby
def namespaced_name
  [namespace, name.camelize].flatten.map(&:camelize).join('::').singularize.constantize
rescue NameError
  name
end
```

So instead of `V2::Invoice` it returns an `Invoice`.

This PR is about fixing that problem and get rid of `rescue` from almost base error class.

### DIFF
--- a/lib/cancan/controller_resource.rb
+++ b/lib/cancan/controller_resource.rb
@@ -262,9 +262,7 @@ module CanCan
     end
 
     def namespaced_name
-      [namespace, name.camelize].flatten.map(&:camelize).join('::').singularize.constantize
-    rescue NameError
-      name
+      ([namespace, name] * '/').singularize.camelize.safe_constantize || name
     end
 
     def name_from_controller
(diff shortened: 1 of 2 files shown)


bambycha:
@Senjai would you like to share any thought about it? Can I be somehow helpful to process merge of this patch?

From my point of view, currently cancancan allows us to pass `@name` as a symbol value. And then trying to invoke `ActiveSupport::Inflector#camelize` on it. That surely fails because AS doesn't extend Symbol class. Yes, we have a `rescue` there but it default some logic. And generally afaik it's a bad practice to use `rescue` in such kind of places.

I have some concerns about using `safe_constantize` because it's availble only for `rails >= 3.2.1`. And you may want to support versions under it as well but I dunno.

Senjai:
@bambycha Thank you, I apologize for the delay in getting to this.

PR #276, opened 2015-11-19
Use `cover?` for ranges

allenwq:
Both `Time` and `ActiveSupport::TimeWithZone` are not iterable:

```
time_range = Time.now..1.day.from_now
time_range.include?(Time.now)

=> TypeError: can't iterate from ActiveSupport::TimeWithZone
```

We should use `cover?` in this case.

Any comments ?

### DIFF
--- a/lib/cancan/rule.rb
+++ b/lib/cancan/rule.rb
@@ -140,6 +140,7 @@ module CanCan
       case value
       when Hash       then hash_condition_match?(attribute, value)
       when String     then attribute == value
+      when Range      then value.cover?(attribute)
       when Enumerable then value.include?(attribute)
       else attribute == value
       end
--- a/spec/cancan/ability_spec.rb
+++ b/spec/cancan/ability_spec.rb
@@ -294,6 +294,13 @@ describe CanCan::Ability do
     expect(@ability.can?(:read, 4..40)).to be(false)
   end
 
+  it "allows a range of time in conditions hash" do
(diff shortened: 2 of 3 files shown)


lib/cancan/rule.rb
```diff
@@ -140,7 +140,7 @@ def condition_match?(attribute, value)
       case value
       when Hash       then hash_condition_match?(attribute, value)
       when String     then attribute == value
-      when Enumerable then value.include?(attribute)
+      when Enumerable then enumerable_condition_match?(attribute, value)
```
lowjoel: I would suggest having a separate case for checking for `Range`

allenwq: I agree, and `range.cover?` also have much better performance than `enumerable.include?`

Senjai:
Looks fine to me code wise. Curious, what do you do in your application that uses this mechanic?

allenwq:
@Senjai We have some abilities like:

`can :attend, Course, end_at: Time.utc(0)..Time.now`

Senjai:
Awesome, is there any chance I could bug you for at least one active record spec that tests this with a model, just to make sure there aren't any regressions?

allenwq:
Sure thing, I will add model test then !

lowjoel:
Thanks @Senjai!

allenwq:
@Senjai Tests added

allenwq:
@Senjai Got anytime to look at this ?

Senjai:
Thanks!

PR #279, opened 2015-12-13
Added "_" for unused variables.

Ashviniv:


lib/cancan/ability.rb
```diff
@@ -322,11 +322,11 @@ def expanded_actions
...
-      subject = if subject.kind_of?(Hash) && subject.key?(:any)
-        subject[:any]
-      else
-        [subject]
-      end
+      _subject = if subject.kind_of?(Hash) && subject.key?(:any)
+                   subject[:any]
+                 else
+                   [subject]
+                 end
```
coorasse: don't rename it. remove it directly.

Senjai:
Thanks

PR #281, opened 2015-12-22
AC::Parameters no longer inherits from HWIA

craig1410:
ActionController::Parameters no longer inherits from HashWithIndirectAccess and so does not respond to :include? It does delegate the has_key? method which does the same job.


craig1410:
I discovered this problem using cancancan on Rails 5.0.0.Beta1 
Looks like this commit in rails led to the issue: https://github.com/rails/rails/commit/14a3bd520dd4bbf1247fd3e0071b59c02c115ce0

PR #282, opened 2015-12-24
This is not the master branch, and 1.13 is latest stable.

will-in-wi:
Noticed that the 1.13.1 tag had a Readme which said it was the master branch. Also, while `~> 1.10` works fine, it doesn't make the user use a more recent version.

This PR updates both.

Thanks!


kimrgrey:
Looks reasonable to me :+1:

Senjai:
Thank you sir.

will-in-wi:
Thank you! And thank you for all of your work on this project!

PR #289, opened 2016-01-18
Rails 5 Tests and Fixes

craig1410:
Sharing the work completed so far on getting cancancan working under Rails 5 and Ruby 2.2.x. 

Currently 9 test failures - starting to work on fixing them. Collaboration welcome.
```
rspec ./spec/cancan/model_adapters/active_record_adapter_spec.rb:162 # CanCan::ModelAdapters::ActiveRecordAdapter should only read articles which are published or in visible categories
rspec ./spec/cancan/model_adapters/active_record_adapter_spec.rb:267 # CanCan::ModelAdapters::ActiveRecordAdapter returns `not (sql)` for single `cannot` definition in front of default `can` condition
rspec ./spec/cancan/model_adapters/active_record_adapter_spec.rb:187 # CanCan::ModelAdapters::ActiveRecordAdapter allows conditions in SQL and merge with hash conditions
rspec ./spec/cancan/model_adapters/active_record_adapter_spec.rb:293 # CanCan::ModelAdapters::ActiveRecordAdapter does not forget conditions when calling with SQL string
rspec ./spec/cancan/model_adapters/active_record_adapter_spec.rb:125 # CanCan::ModelAdapters::ActiveRecordAdapter fetches any articles which are published or secret
rspec ./spec/cancan/model_adapters/active_record_adapter_spec.rb:145 # CanCan::ModelAdapters::ActiveRecordAdapter fetches only the articles that are published and not secret
rspec ./spec/cancan/model_adapters/active_record_adapter_spec.rb:135 # CanCan::ModelAdapters::ActiveRecordAdapter fetches any articles which we are cited in
rspec ./spec/cancan/model_adapters/active_record_adapter_spec.rb:249 # CanCan::ModelAdapters::ActiveRecordAdapter returns SQL for single `can` definition in front of default `cannot` condition
rspec ./spec/cancan/model_adapters/active_record_adapter_spec.rb:273 # CanCan::ModelAdapters::ActiveRecordAdapter returns appropriate sql conditions in complex case
```


lib/cancan/model_adapters/active_record_5_adapter.rb
```diff
@@ -0,0 +1,60 @@
...
+          table = @model_class.send(:arel_table)
+          table_metadata = ActiveRecord::TableMetadata.new(@model_class, table)
+          predicate_builder = ActiveRecord::PredicateBuilder.new(table_metadata)
+
+          conditions = predicate_builder.resolve_column_aliases(conditions)
+          conditions = @model_class.send(:expand_hash_conditions_for_aggregates, conditions)
+
+          conditions.stringify_keys!
+
+          conditions, binds = predicate_builder.create_binds(conditions)
```
Dahlgren: By removing this line the rest of the failing tests are resolved

craig1410:
Many thanks to @Dahlgren for pointing out the simple fix required to get tests passing. Very much appreciated! :smile: 
Updated for beta 2 also.

lib/cancan/controller_additions.rb
```diff
@@ -12,7 +12,7 @@ module ClassMethods
       #   end
       #
       def load_and_authorize_resource(*args)
-        cancan_resource_class.add_before_filter(self, :load_and_authorize_resource, *args)
+        cancan_resource_class.add_before_action(self, :load_and_authorize_resource, *args)
```
Senjai: before_action is not defined in rails 3.2. https://github.com/rails/rails/blob/3-2-stable/actionpack/lib/abstract_controller/callbacks.rb

I'm concerned if this passes tests. Or I might be missing something here.

craig1410: Apologies for delay in responding @Senjai , I thought I had already done so. Yeah good spot there and yes it is worrying that it passes tests. Do you have any suggestions on how I might use *_filter in Rails 3.2 and *_action in Rails 4+. Is it just case of having loads of conditionals throughout the code or is there a better way? I don't mind doing the work.

Ziaw: _filter calls can be abstracted with model_adapters/active_record_N_adaper.rb

alejorivera:
What's missing for this to be merged? How can I help?

craig1410:
@alejorivera I think all that's left is to figure out why the Rails 3.2 tests are passing when they shouldn't and then to create the abstraction layer to allow the Rails 3.2 *_filter calls to be used for Rails 3.2 and the *_action calls to be used for later versions. I just can't seem to find time to devote to this at the moment. If this is something you feel you could tackle then it would be greatly appreciated. If not then I'll try to get around to it asap. Let me know if you decide to work on it so we don't end up duplicating effort.

Many Thanks,
Craig.

alejorivera:
@craig1410 Hey! Thanks for the quick response. I'll do my best but I've never contributed to open-source before, so I don't know if my efforts will be helpful!

craig1410:
@alejorivera I know the feeling. I've been doing commercial Rails development for over 5 years but sadly I've never really contributed much OSS back until recently. I'm trying to make up for that a bit now but it's hard to fit in the day job at the same time as doing side projects and learning about new stuff and still have time left over to contribute back. I find it especially hard as a contractor where it's expected that everything I do is of direct value to the business who is hiring me. They often don't understand the huge benefits which come from OSS.

Getting back to the pull request, I might get some time to look at it this week and if I do then I'll post an update here. If you do the same then we should be able to minimise duplication of effort. That said, it's no bad think if we end up with two solutions as we can just let those with merge rights pick the one they like best. :)

kartikluke:
Rails 5.0.0rc1 is out. You might want to consider that instead of beta2 or will it automatically update to the latest release.

craig1410:
@kartikluke yeah, I'll update it to reference RC1. I knew Beta 4 was out but hadn't spotted the RC1 release so thanks for the heads-up! One step closer to Rails 5 release so I'll do my best to spend some time on the above this week.

yjukaku:
So just to clarify:

1. Do the controller spec needs to be split into pre-Rails5 and Rails5+ so that the `expect(@controller_class).to receive(:before_action).with...` calls can be tested properly?
2. There needs to be an abstraction similar to the active_record_N_adapter, but for controllers?

craig1410:
@yjukaku yes I think that is correct. I tried looking at point 2 the other night but didn't get very far as I'm not that familiar with the cancancan codebase yet. I think I understand what needs to be done but I'm not sure exactly where to insert the abstraction. If you are able to help in this area then that would be very welcome.

yjukaku:
@craig1410 Your last commit was what I just sat down to do 👍 👍 

I'll start looking at point 2 now, but I'm also unfamiliar with the `cancancan` code base.

craig1410:
Merged in latest from the develop branch so we now have failing tests for Rails 3.2. All we need now is an ActionController abstraction to use the appropriate callbacks for pre or post Rails 4. Suggestions on how to do this are most welcome.

yjukaku:
This is not a pure abstraction but likely would work:

```ruby
# at the top of  lib/cancan/controller_resource.rb
    def self.add_before_action(controller_class, method, *args)
      options = args.extract_options!
      resource_name = args.first
      before_action_method = before_callback_name(controller_class, options.delete(:prepend))
      controller_class.send(before_action_method, options.slice(:only, :except, :if, :unless)) do |controller|
        controller.class.cancan_resource_class.new(controller, resource_name, options.except(:only, :except, :if, :unless)).send(method)
      end
    end

    def self.before_callback_name(controller_class, prepend)
      if prepend
        controller_class.respond_to?(:prepend_before_action) ? :prepend_before_action : :prepend_before_filter
      else
        controller_class.respond_to?(:before_action) ? :before_action : :before_filter
      end
    end
```

For some reason all of my tests are failing when I pull down the latest repo so can't test this.

craig1410:
@yjukaku how are you running the tests? Are you using appraisal rake? If so then it will fail at the first appraisal because Rails 3.2 is the first appraisal. Make sure you've done bundle install then appraisal install first. Then you can try each appraisal like this: appraisal activerecord_5.0 rake

craig1410:
So, failing tests created and fixes applied to make tests pass. I'm sure there is a more elegant way to abstract this out but hopefully this solution will be acceptable in the meantime.

lib/cancan/model_adapters/active_record_5_adapter.rb
```diff
@@ -0,0 +1,58 @@
+module CanCan
+  module ModelAdapters
+    class ActiveRecord5Adapter < AbstractAdapter
```
Senjai: Are there really any differences between the active record 4 adapter and this one? Besides removing version checks?

If we have to keep this as a seperate file, it needs its own tests. If possible though (and if they are as similar as I presume) I would like to make the active record 4 adapter work for both major versions. Unless there is a reason to presume otherwise. I may be missing something here so feel free to enlighten me!

craig1410: They are almost identical apart from the version checks with the exception of the sanitize_sql method which has a slightly different structure around the predicate builder. To be honest it was a few months ago that I looked at this originally and I can't recall the exact reasons for those changes.

yjukaku:
Would it be acceptable to just write `controller.respond_to? ? :before_action : :before_filter` etc. where appropriate?

Senjai:
@craig1410 There are a decent amount of nitpicks I'd like to address here, but overall this looks fine to get this moving forward. Do you have any concerns with backwards compatibility?

craig1410:
@Senjai I'm happy to help out sorting the nitpicks of course. I don't have any specific backwards compatibility concerns other than that support for mongoid and sequel seem to be falling behind activerecord a bit. Do you want me to look at consolidating the AR4 and AR5 adapters?

Senjai:
@craig1410 If you had the time, I would appreciate that. As they aren't different, save the active record version checks really.

craig1410:
@Senjai not a problem, new commit pushed, please let me know if you need anything else fixed on the pull request.

craig1410:
@Senjai anything else you need me to do to this pull request before it can be merged? I've probably got some time available in the next day or two if required.

Senjai:
@craig1410 This looks good to me. Once rails 5 is actually out I'll axe rails 3.2 support and we'll be able to cut all the callback_action stuff.

lib/cancan/model_adapters/active_record_5_adapter.rb
```diff
@@ -46,8 +46,6 @@ def sanitize_sql(conditions)
 
           conditions.stringify_keys!
 
-          conditions, binds = predicate_builder.create_binds(conditions)
```
Senjai: Just for curiosity sake, why was this here in the first place?

craig1410: Quite honestly, I'm not sure now. At the time I was trying to make sense of the changes in Rails 5 and was digging deep into source code to get things working, specifically around the predicate builder. I got things mostly working but another contributor @dahlgren discovered that removing this line made the tests pass. Sorry I can't provide a better explanation.

craig1410:
@Senjai point taken re the commit messages, will bear that in mind in future. Thanks! :)

atussec:
is the pull request merge to master branch?

Senjai:
@atussec @craig1410 Released in 1.15. Thanks for the hard work here.

PR #292, opened 2016-01-28
use `ActiveSupport.on_load` hook to include `CanCan::ModelAdditions` in AR::Base

kuldeepaggarwal:


Senjai:
:+1: on this, but it doesnt seem to work for 3.2.

kuldeepaggarwal:
Build is not failing because of this change. I am sure build is already broken on master.

Senjai:
@kuldeepaggarwal I don't think so: https://travis-ci.org/CanCanCommunity/cancancan/builds

kuldeepaggarwal:
@Senjai Updated the PR.

kuldeepaggarwal:
@Senjai: I tested the code locally. Test is showing very weird behaviour, sometimes they are passing and sometimes getting failed.

Senjai:
It looks like it's a load order issue then. Odds are the model we're including (ModelAdditions) is being included before other modules that define constants our model uses. We should be able to get around this by using an initializer instead of just monkeypatching the model when the file is required.

gsmetal:
Any update here? I can't deploy my application to Dokku because assets:precompile failing without database on build phase (https://devcenter.heroku.com/articles/rails-asset-pipeline#troubleshooting).

gsmetal:
If anyone interested, there is workaround for this: `require: false` in Gemfile and initializer with: 
```ruby
ActiveSupport.on_load(:active_record) do
  require 'cancancan'
end
```

alexsanford:
Oh my. I just submitted a patch that does the same thing (344). Can't believe I didn't see this first! 😫 

Should we try to resolve this? Or should we just drop support for 3.2? It's now an unsupported version of Rails, so I would be in favor of dropping support for it and merging in this PR.

coorasse:
Since 3.2 is not supported anymore I will schedule this PR for version 2.0 and drop support for ActiveRecord 3.2

kuldeepaggarwal:
Thanks @coorasse

PR #296, opened 2016-02-02
Build the join and used the assigned table name for building the conditions hash

### LINKED ISSUE #284: Table aliases not are used in WHERE clause
I am working on a gamified learning platform where students in courses can earn experience points and am trying to allow admins of a particular course to view `ExperiencePointsRecord`s of students from that same course, but ran into Coursemology/coursemology2#706.

When a table has to be joined twice, the second instance of the table in aliased is the `FROM` clause, but the conditions in the `WHERE` clause does not use the alias. 

Models

``` ruby
def Course
    has_many :course_users
end

def User
    has_many :course_users
end

def CourseUser
    belongs_to :user
    belongs_to :course
end

def ExperiencePointsRecord
    belongs_to :course_user
end
```

Ability definition

``` ruby
read_hash = { course_user: { course: { course_users: { user_id: user.id, role: 'admin' } } } }
can :read, ExperiencePointsRecord, read_hash
```

Calling `accessible_by` generates a query like:

``` sql
...
FROM
    "experience_points_records" 
        LEFT OUTER JOIN "course_users"
            ON "course_users" . "id" = "experience_points_records" . "course_user_id" 
        LEFT OUTER JOIN "courses"
            ON "courses" . "id" = "course_users" . "course_id"
                AND "courses" . "instance_id" = $1 
        LEFT OUTER JOIN "course_users" "course_users_courses"
            ON "course_users_courses" . "course_id" = "courses" . "id"
WHERE
    "experience_points_records" . "course_user_id" = $2
    AND (
        "course_users" . "user_id" = $3
        AND "course_users" . "role" = $4
    )  [["instance_id", 1], ["course_user_id", 4], ["user_id", 9], ["role", "admin"]]
```

which is incorrect.  `course_users` should be generated as `course_users_courses` in the WHERE clause:

``` sql
WHERE
    "experience_points_records" . "course_user_id" = $2
    AND (
        "course_users_courses" . "user_id" = $3
        AND "course_users_courses" . "role" = $4
    )  [["instance_id", 1], ["course_user_id", 4], ["user_id", 9], ["role", "admin"]]
```

Is this a bug?

lowjoel:
This fixes #284.

What this does is, instead of assuming that tables always get the same SQL table name, we actually perform the join to figure out what alias the table is given, and use that to specify the conditions.

cc @kxmbrian

### DIFF
--- a/lib/cancan/model_adapters/active_record_3_adapter.rb
+++ b/lib/cancan/model_adapters/active_record_3_adapter.rb
@@ -6,6 +6,29 @@ module CanCan
         model_class <= ActiveRecord::Base
       end
 
+      def tableized_conditions(conditions, model_class = @model_class)
+        return conditions unless conditions.kind_of? Hash
+        conditions.inject({}) do |result_hash, (name, value)|
+          if value.kind_of? Hash
+            value = value.dup
+            association_class = model_class.reflect_on_association(name).klass.name.constantize
+            nested = value.inject({}) do |nested,(k,v)|
+              if v.kind_of? Hash
+                value.delete(k)
+                nested[k] = v
+              else
+                result_hash[model_class.reflect_on_association(name).table_name.to_sym] = value
+              end
+              nested
+            end
+            result_hash.merge!(tableized_conditions(nested,association_class))
+          else
+            result_hash[name] = value
+          end
+          result_hash
+        end
+      end
+
       private
 
       def build_relation(*where_conditions)
--- a/lib/cancan/model_adapters/active_record_4_adapter.rb
+++ b/lib/cancan/model_adapters/active_record_4_adapter.rb
@@ -6,28 +6,27 @@ module CanCan
         model_class <= ActiveRecord::Base
       end
 
-      def tableized_conditions(conditions, model_class = @model_class,
-                               current_nesting = [], current_scope = model_class.all)
+      def tableized_conditions(conditions, current_scope = @model_class.all, current_nesting = [],
+                               table_aliases = {})
         return conditions unless conditions.kind_of? Hash
         conditions.inject({}) do |result_hash, (name, value)|
           if value.kind_of? Hash
-            value = value.dup
             new_nesting = current_nesting + [name]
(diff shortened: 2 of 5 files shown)


lowjoel:
@Senjai seems like AR3.2 doesn't like this... do we make this AR4 only?

lowjoel:
I've made it AR4 only. Let me know after review, I'll squash the commits.

lowjoel:
#238 might allow us to deduce the table names more precisely, but until that's merged, I'll assume joins and includes generates the same thing.

lowjoel:
My latest commit actually builds the joins, and then asks AR to build the join dependency tree, then I get the aliases from the dependency tree. This is about as accurate as it gets, IMO.

lowjoel:
...but this doesn't work with Rails 4.0. What do you recommend I do?

Senjai:
Hi @lowjoel. I've given this a cursory review. I think we're running into a lot of problems with our adapters when we try to work around these scenarios. We want to have an interface that operates like the interface we use with active record to perform queries, and we want that interface because of how easy it is to define rules using it.

The problem in my experience is, that this couples us quite extensively to the internal nuances and quirks of each adapter, and their versions. This adds quite a bit of complexity and I'm not sure bringing it in as is is ideal at this moment. I'm not sure how we're approaching this problem is the correct way of doing so. We're expending a lot of effort to recreate an interface that we should just be able to use.

lowjoel:
@Senjai exactly -- the logic here has been solved by Squeel, so I'm basically just reimplementing parts of their replacements for AR and Arel. Let me know what the future direction is and I'll see how to help.

Meanwhile I'm using this in my Rails 4.2 app, looks fine at the moment.

lowjoel:
I think it's cleaner to just use squeel directly, see #307.

FWMatt:
+1 This fixed a nasty issue for me where conditions were overwriting each other.

coorasse:
I think the basecode changed too much since this PR was opened. Please feel free to reopen based on the current version

PR #302, opened 2016-03-10
Safe constatize to avoid uninitialized constant errors

teosz:


teosz:
@Senjai ? :(

coorasse:
Why you want to make a `safe_constantize`? It will simply return nil instead of throwing an exception...but from my point of view raising an exception is the right behaviour here. Do you have an example where this may be useful? Thanks.

coorasse:
I close due to a lack of feedback. Feel free to reopen

PR #308, opened 2016-03-29
docs: update branch changes in README

raszi:


coorasse:
This PR would bring conflicts when we merge develop into master....if we are following gitflow...I don't knwo... @Senjai ?

raszi:
Right now the main page has broken references.

As right now according to [this link](http://stackoverflow.com/a/18677732/221213) it is not possible to use the branch name in the README, so either the conflicts or the broken references.

Senjai:
I do want to switch to using the master branch. The decision to use develop was made before I was involved. The problem is that the existing PR's against develop cannot be retargeted so I'd have to keep them in sync for a while after anyway.

Senjai:
I do want to switch to using the master branch. The decision to use develop was made before I was involved. The problem is that the existing PR's against develop cannot be retargeted so I'd have to keep them in sync for a while after anyway.

Senjai:
I do want to switch to using the master branch. The decision to use develop was made before I was involved. The problem is that the existing PR's against develop cannot be retargeted so I'd have to keep them in sync for a while after anyway.

PR #314, opened 2016-04-12
Avoid unnecessary checks [develop]

glebm:
* Evaluate `:all` rules before any other rules in the consecutive `base_behavior == true` rules.
* Abort evaluation on the first matching subject.
* Use JRuby 9 on Travis (for `Enumerable#lazy`).

Also backported to 1.13.x in #315.

### DIFF
--- a/lib/cancan/ability.rb
+++ b/lib/cancan/ability.rb
@@ -61,14 +61,11 @@ module CanCan
     #
     # Also see the RSpec Matchers to aid in testing.
     def can?(action, subject, *extra_args)
-      subject = extract_subjects(subject)
-
-      match = subject.map do |subject|
-        relevant_rules_for_match(action, subject).detect do |rule|
-          rule.matches_conditions?(action, subject, extra_args)
+      match = extract_subjects(subject).lazy.map do |a_subject|
+        relevant_rules_for_match(action, a_subject).detect do |rule|
+          rule.matches_conditions?(action, a_subject, extra_args)
         end
-      end.compact.first
-
+      end.reject(&:nil?).first
       match ? match.base_behavior : false
     end
     # Convenience method which works the same as "can?" but returns the opposite value.
@@ -369,7 +366,22 @@ module CanCan
         rule.relevant? action, subject
       end
       relevant.reverse!.uniq!
+      optimize_order! relevant
       relevant
+    end
+
+    # Optimizes the order of the rules, so that rules with the :all subject are evaluated first.
+    def optimize_order!(rules)
+      first_can_in_group = -1
+      rules.each_with_index do |rule, i|
+        (first_can_in_group = -1) and next unless rule.base_behavior
+        (first_can_in_group = i) and next if first_can_in_group == -1
+        if rule.subjects == [:all]
+          rules[i] = rules[first_can_in_group]
+          rules[first_can_in_group] = rule
+          first_can_in_group += 1
+        end
+      end
     end
 
     def possible_relevant_rules(subject)
--- a/spec/cancan/ability_spec.rb
+++ b/spec/cancan/ability_spec.rb
@@ -52,6 +52,22 @@ describe CanCan::Ability do
     expect(@ability.can?(:read, 6)).to be(true)
   end
 
+  it "performs can(_, :all) before other checks when can(_, :all) is defined before" do
+    @ability.can :manage, :all
+    @ability.can :edit, String do |_string|
+      fail 'Performed a check for :edit before the check for :all'
+    end
+    expect { @ability.can? :edit, 'a' }.to_not raise_exception
+  end
+
+  it "performs can(_, :all) before other checks when can(_, :all) is defined after" do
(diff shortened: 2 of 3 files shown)


lib/cancan/ability.rb
```diff
@@ -61,15 +61,12 @@ module Ability
     #
     # Also see the RSpec Matchers to aid in testing.
     def can?(action, subject, *extra_args)
-      subject = extract_subjects(subject)
-
-      match = subject.map do |subject|
-        relevant_rules_for_match(action, subject).detect do |rule|
-          rule.matches_conditions?(action, subject, extra_args)
+      extract_subjects(subject).each do |a_subject|
```
coorasse: to me this looks like a refactoring which does not add anything relevant or improve performances. Am I wrong?

lib/cancan/ability.rb
```diff
@@ -369,9 +366,30 @@ def relevant_rules(action, subject)
         rule.relevant? action, subject
       end
       relevant.reverse!.uniq!
+      optimize_order! relevant
       relevant
     end
 
+    # Optimizes the order of the rules, so that rules with the :all subject are evaluated first.
+    def optimize_order!(rules)
+      fst_pos_i = -1
```
coorasse: this variable name is not significant. is it `first_position_index`?

lib/cancan/ability.rb
```diff
@@ -369,9 +366,30 @@ def relevant_rules(action, subject)
...
+      rules.each_with_index do |rule, i|
+        unless rule.base_behavior
```
coorasse: can we inline this doing something like `first_position_index = -1 and next unless rule.base_behaviour`?

lib/cancan/ability.rb
```diff
@@ -369,9 +366,30 @@ def relevant_rules(action, subject)
...
+          fst_pos_i = -1
+          next
+        end
+        if fst_pos_i == -1
```
coorasse: same here. `first_position_index = i and next if first_position_index == -1`

lib/cancan/ability.rb
```diff
@@ -369,9 +366,30 @@ def relevant_rules(action, subject)
         rule.relevant? action, subject
       end
       relevant.reverse!.uniq!
+      optimize_order! relevant
       relevant
     end
 
+    # Optimizes the order of the rules, so that rules with the :all subject are evaluated first.
+    def optimize_order!(rules)
```
coorasse: does something like 
```ruby
rules.sort_by { |rule| rule.subjects == [:all] ? 0 : 1}
```

return the same result?

coorasse:
Very nice idea @glebm! I like it very much. But I also have some questions as you can see. Thanks for your effort!

lib/cancan/ability.rb
glebm: I think it will now abort at the first matching subject, whereas previously it would find one matching rule for each subject.

coorasse: `detect` stops on the first matching result already (http://ruby-doc.org/core-2.3.0/Enumerable.html#method-i-detect)

lib/cancan/ability.rb
glebm: It won't. This code puts `[:all]` rules before other rules within consecutive groups of `base_behavior == true` rules. We cannot reorder `true` and `false` clause relative to each other because they are not disjuntive (the order matters, `cannot` takes priority over a previously defined `can`).

lib/cancan/ability.rb
glebm: Done, with some clarifying parenthesis.

lib/cancan/ability.rb
glebm: Done

lib/cancan/ability.rb
glebm: Renamed to `first_can_in_group`

lib/cancan/ability.rb
coorasse: :+1:

spec/cancan/ability_spec.rb
```diff
@@ -52,6 +52,22 @@
     expect(@ability.can?(:read, 6)).to be(true)
   end
 
+  it "does not perform unnecessary checks" do
```
coorasse: nice tests

coorasse:
I can't merge it but you have my ok. Thanks again. :+1:

lib/cancan/ability.rb
glebm: `detect` does, but not the outer `.map` loop. Refactored again to use `lazy` instead of `return`, so it's cleaner now.

glebm:
@coorasse I've also backported it to 1.13.x in #315. Who can I ping to merge this? A bit concerned since there has been no commits since Feb 29th.

coorasse:
@Senjai

Senjai:
:+1: Thank you @glebm

PR #315, opened 2016-04-14
Avoid unnecessary checks [1.13.x-maintenance]

glebm:
* Evaluate `:all` rules before any other rules in the consecutive `base_behavior == true` rules.
* Abort evaluation on the first matching subject.
* Use JRuby 9 on Travis (for `Enumerable#lazy`).


Senjai:
No need for this one, I'll cut a release for a new 1.13 after work today. You don't have to send patches to seperate versions :P

PR #316, opened 2016-04-15
Controller asks current_ability if it has been authorized

houen:
In our company we want to move the `authorize!` calls out of our controllers. Unfortunately this currently means we would have to give up `check_authorization`, since this currently relies on the instance var @_authorized being set on the controller.

This PR changes that so the controller asks the current_ability if it has had `authorize!` called. Now we can pass the current ability into other objects, and still use `check_authorization` in our controllers.

I did not know what to add to the CHANGELOG. Any help is appreciated.

### DIFF
--- a/lib/cancan/ability.rb
+++ b/lib/cancan/ability.rb
@@ -222,7 +222,7 @@ module CanCan
 
     # Returns whether this ability has had authorize! called
     def authorize_called?
-      @_authorized
+      @_authorized == true
     end
 
     def unauthorized_message(action, subject)
--- a/lib/cancan/ability.rb
+++ b/lib/cancan/ability.rb
@@ -207,6 +207,8 @@ module CanCan
 
     # See ControllerAdditions#authorize! for documentation.
     def authorize!(action, subject, *args)
+      @_authorized = true
+
       message = nil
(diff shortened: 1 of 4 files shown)


Senjai:
Does it make more sense to deprecate check_authorization? I dont know why that should be there when we really just want to explicitly authorize resources.

houen:
@Senjai I would not deprecate functionality others are relying on. There is essentially much good in check_authorization. It is a good guard against accidentally missing authorization. 

You could home-roll it of course, but I think many people using cancancan are not at that level. 

When I first started with cancan back in the day, it was a great help. Rather, I would merge this PR 😃

houen:
Oh, and thank you for taking care of my duplicates :)

Senjai:
I'm not sure about this one. It's rolled into the controller additions model, but I've never used check_authorization. Simply because I code the authorization into my controllers explicitly, and it seems like more of a tool to prevent developers from forgetting to call authorize in certain controllers. 

I would almost prefer this to be handrolled in applications that require this behaviour as I don't think it's a practice we should encourage in cancancan. Expecting that "something" has been `authorize!`d has no value if it doesn't check that the correct "thing" has been authorized.

coorasse:
+1 for deprecating `check_authorization`

Issue #304, opened 2016-03-23
Add error class method-pattern


shlima:
Please add method `unauthorized_error(action, subject)` by analogy with `unauthorized_message` https://github.com/CanCanCommunity/cancancan/blob/develop/lib/cancan/ability.rb#L217-L219 

In this case I can override this method to raise PaymentRequiredError instead of AccessDeniedError for some actions for example

PR #319, opened 2016-04-21
Allow authorize! to use a custom error class

neeaagh:
This is in reference to #304. 

You can give `authorize!` an error class other than the default of `CanCan::AccessDenied` with:

```ruby
class PaymentRequiredError < CanCan::AccessDenied; end

authorize! :read, :foo, message: 'Payment method required!', error: PaymentRequiredError
```

### DIFF
--- a/spec/cancan/ability_spec.rb
+++ b/spec/cancan/ability_spec.rb
@@ -419,7 +419,7 @@ describe CanCan::Ability do
     expect(@ability.attributes_for(:new, Range)).to eq({:foo => "foo", :bar => 123, :baz => "baz"})
   end
 
-  it "raises access denied exception if ability us unauthorized to perform a certain action" do
+  it "raises access denied exception if ability is unauthorized to perform a certain action" do
     begin
       @ability.authorize! :read, :foo, 1, 2, 3, :message => "Access denied!"
     rescue CanCan::AccessDenied => e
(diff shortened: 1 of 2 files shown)


spec/cancan/ability_spec.rb
```diff
@@ -431,6 +431,20 @@ class Container < Hash; end
     end
   end
 
+  it "raises a custom exception if ability is unauthorized to perform a certain action and is given a custom error class" do
```
coorasse: I don't like this test very much. I'd like more to have one test for each expectation and to actually check that `is_expected.to raise_error(MyCustomAccessDenied)` https://www.relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/raise-error-matcher

neeaagh:
Thanks for the quick feedback!

Would you be looking for something more like this?

```ruby
  context "when using a custom error class" do
    class MyCustomAccessDenied < CanCan::AccessDenied; end

    it "raises a custom exception if ability is unauthorized to perform a certain action" do
      expect{ @ability.authorize! :read, :foo, 1, 2, 3, message: "Access denied!", error: MyCustomAccessDenied }
        .to raise_error(MyCustomAccessDenied)
    end

    it "maintains properties of default error class" do
      begin
        @ability.authorize! :read, :foo, 1, 2, 3, message: "Access denied!", error: MyCustomAccessDenied
      rescue MyCustomAccessDenied => e
        expect(e.message).to eq("Access denied!")
        expect(e.action).to eq(:read)
        expect(e.subject).to eq(:foo)
      else
        fail "Expected MyCustomAccessDenied exception to be raised"
      end
    end
  end
```

Do you think checking for the message, action and subject is necessary or is the `raise_error` sufficient?

Senjai:
I'm always hesitant to add code like this into cancancan at this level. This is because controllers, at the time of using `authorize!` can `rescue_from` or `resuce CanCan::AccessDenied` with a message of their choosing. They can then re raise the message, or add logic as they see fit there so I'm hesitant to expand the interface to accommodate it.

For something like a `PaymentRequiredError` that isn't even authorization related at all. That comes down to your application logic and you shouldn't be checking these kind of things at the authorization level.

PR #330, opened 2016-06-21
Changed includes to joins so .select() can still be used

akuma12:
I noticed that .select() cannot be used with a cancan loaded query, as .includes() actually renames the columns. If .includes() is changed to .joins(), you get the same result with the ability to use custom selects.


Senjai:
You dont get the same result. This is a hard problem a lot of people are trying to tackle.

Using joins performs an inner join, and thus can't be used with queries or logic that expect something similar to an outer join.

coorasse:
You may want to contribute to this: https://github.com/CanCanCommunity/cancancan/pull/238

Issue #286, opened 2016-01-05
Rails 5 Support


craig1410:
Just wondering if anyone is working on Rails 5 support? I've spent quite a bit of time looking at various Rails 5 related issues today and have made some progress but I've gotten a bit bogged down on the sanitise_sql method in the active record adapter. Before I spend any more time on this I thought I'd ask what the official status is?

kimrgrey:
Hi, @craig1410!

It looks like there is no "official" status :smirk: I'll try to contact some contributors to find out who is currently working on the project and what's a plan. But anyway it will good to see Rails 5 support draft. 

And, please, do not forget to share your experience not just your code! With what problems did you face? How did you solve it? And so on.

Senjai:
Yes we will be looking at rails 5 support. Pull requests always welcome, and I know there are a few outstanding. Unfortunatley our industry is famous for deadlines and I have a large one on the 15th, I'll be free after then.

craig1410:
Thanks guys, good to hear that it's on the agenda. I just didn't want to spend more time on this in case I was just duplicating effort. It doesn't sound like anyone has an imminent rails 5 release so I don't mind continuing with it.

I'll see if I can organise what I've already done into a pull request consisting of a series of self contained commits so that you can either accept the PR as a whole with amendments as required or just cherry-pick individual commits.

@Senjai I hope your impending deadline passes successfully.

@kimrgrey I take on board what you said about sharing experience etc. I'll try to document my Rails 5 experience as I organise my code for the PR.

adityashedge:
I want to use 'cancancan' with Rails 5 beta 1 in production and wanted to know about the known issues.
If we list out the issues then I can also contribute to the fixes.

craig1410:
I've posted a provisional PR and will work on fixing the failing tests. Feel free to assist, cherry-pick or ignore as appropriate. PR #289

Senjai:
Closed per #289

PR #333, opened 2016-07-13
Drop support for Rails < 4.2

oliverklee:
According to http://guides.rubyonrails.org/maintenance_policy.html , Rails
3.2, 4.0 and 4.1 have reached their end of life (as Rails 5.0 has been
released).  So we stop supporting it, allowing for faster builds on Travis,
code removal and simpler code.


tarunjadhwani:
So cancancan works fine with rails 5, right?

oliverklee:
I don't know. I haven't tested it yet (manually), there's still issue #286 for this, and the Rails 5 builds are still excluded from the build matrix: https://github.com/CanCanCommunity/cancancan/blob/develop/.travis.yml#L22

It looks like some Rails 5 work was done recently, though.

Senjai:
Thank you for this. I'll be including it in the next major.

coorasse:
I think is time for this to be merged. 
Maintenance for Rails 3.2 has been dropped now.
I would start a cancancan 2.0 version with this merged.

PR #344, opened 2016-08-16
Fix issue with ActiveRecord initialization

### LINKED ISSUE #342: Cancancan breaks belongs_to_required_by_default in Rails 5
In Rails 5 we have

> **#belongs_to is required by default**
> From now on every Rails application will have a new configuration option config.active_record.belongs_to_required_by_default = true, it will trigger a validation error when trying to save a model where belongs_to associations are not present.

I have cancancan 1.15.0 and it breaks default behavior so I still can save model with nil relation.
In case of that I have:

``` ruby
Rails.application.config.active_record.belongs_to_required_by_default
 => true 
ActiveRecord::Base.belongs_to_required_by_default
 => nil 
User.belongs_to_required_by_default
 => nil 
```

Without cancancan included to Gemfile

``` ruby
Rails.application.config.active_record.belongs_to_required_by_default
 => true 
ActiveRecord::Base.belongs_to_required_by_default
 => true 
User.belongs_to_required_by_default
 => true 
```

alexsanford:
Fixes #342

This patch ensures that the model adapter code requiring ActiveRecord is not run until after ActiveRecord has been loaded (explanation [here](https://github.com/CanCanCommunity/cancancan/issues/342#issuecomment-239687074)).

It's not clear to me how I would write a spec for this. I would love to be able to check the actual config values that caused the problem in the original issue. But that would probably involve running a dummy Rails app, which seems like overkill for such a small fix. Thoughts?

### DIFF
--- a/CHANGELOG.rdoc
+++ b/CHANGELOG.rdoc
@@ -2,6 +2,8 @@ Develop
 
 Unreleased
 
+* Fix bug in ActiveRecord configuration (alexsanford)
+
 1.15.0 (June 13th, 2016)
 
 * Add support for Rails 5 (craig1410)
--- a/lib/cancan/model_adapters/active_record_adapter.rb
+++ b/lib/cancan/model_adapters/active_record_adapter.rb
@@ -143,6 +143,8 @@ module CanCan
   end
 end
 
-ActiveRecord::Base.class_eval do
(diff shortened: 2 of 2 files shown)


alexsanford:
Darn! Looks like this is a duplicate of #292. Whoops!

Issue #218, opened 2015-06-17
Wiki - respond_to instead of explicit redirect


akolomiychuk:
I think suggesting using redirect explicitly isn't best option in wiki...

From https://github.com/CanCanCommunity/cancancan/wiki/Exception-Handling :
```ruby
class ApplicationController < ActionController::Base
  rescue_from CanCan::AccessDenied do |exception|
    redirect_to main_app.root_url, :alert => exception.message
  end
end
```

Because there is an issue with ajax requests: if I request a url (to which I don't have access) via ajax I get redirected to the front page inside ajax request and get status 200 OK. This leads to triggering jQuery  ajax success callback, but it shouldn't. I expecting to get `403 Forbidden` instead of `200 OK` in that case.

I think much better will be update wiki to suggest using something like that in order to get correct behavior.

```ruby
rescue_from CanCan::AccessDenied do |exception|
  respond_to do |format|
    format.json { render nothing: true, status: :forbidden }
    format.html { redirect_to main_app.root_url, :alert => exception.message }
    format.js   { render nothing: true, status: :forbidden }
  end
end
```

anthonylewis:
Have you seen the [Authorization in Web Services](https://github.com/CanCanCommunity/cancancan/wiki/Authorization-in-Web-Services) page? I think that would be a good place for your suggestion.

But, ajax requests are so common these days it should probably be mentioned on the [Exception Handing](https://github.com/CanCanCommunity/cancancan/wiki/Exception-Handling) page, too.

Senjai:
I'm fine with either of these. Remember that the example here is to illustrate how to rescue from an AccessDenied exception, not exactly what to do when you do. We should make the assumption the reader knows how to redirect and render based on format in rails, as that is out of scope for this gem.

But I dont mind it being there. Just that we shouldn't focus on it.

anthonylewis:
@Senjai I know from experience that many developers copy and paste from documentation. I agree that this is an example, but I'll bet this exact code is in many apps. I'll update the wiki page in a bit and try to incorporate @akolomiychuk's suggestion.

anthonylewis:
@Senjai @akolomiychuk I updated the two wiki pages mentioned above. Let me know what you think.

akolomiychuk:
Thanks for your feedback. The change is seems pretty well as for me.

[...]

PR #350, opened 2016-09-15
Updates Readme

F3PiX:
- Exception handling: updated, consistent with wiki and #218
- Add 'Rails 5' to instruction for gem file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,9 +23,11 @@ Any help is greatly appreciated, feel free to submit pull-requests or open issue
 
 ## Installation
 
-In **Rails 3 and 4**, add this to your Gemfile and run the `bundle install` command.
+Add this to your Gemfile: 
 
     gem 'cancancan', '~> 1.10'
+    
+and run the `bundle install` command.
 
 ## Getting Started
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -23,7 +23,7 @@ Any help is greatly appreciated, feel free to submit pull-requests or open issue
 
 ## Installation
 
-In **Rails 3 and 4**, add this to your Gemfile and run the `bundle install` command.
+In **Rails 3, 4 and 5 **, add this to your Gemfile and run the `bundle install` command.
```
coorasse: The docs now say that Cancancan supports only Rails >= 3. I think is not necessary to explicitly specify all the versions and not even Rails. Just Leave: `Add this to your Gemfile...`

README.md
```diff
@@ -162,8 +162,12 @@ If the user authorization fails, a `CanCan::AccessDenied` exception will be rais
 ```ruby
 class ApplicationController < ActionController::Base
   rescue_from CanCan::AccessDenied do |exception|
-    redirect_to root_url, :alert => exception.message
-  end
+      respond_to do |format|
+        format.json { head :forbidden, content_type: "text/html" }
+        format.html { redirect_to main_app.root_url, :notice => exception.message }
+        format.js   { head :forbidden, content_type: "text/html"}
```
coorasse: I like the idea of giving an example of all possible formats. Please prefer single quotes over double quotes when  they are not necessary: `"text/html"` --> `'text/html'` and use new hash rocket syntax. `:notice => exception.message` --> `notice: exception.message`

F3PiX: Thnx coorasse, done and done

PR #360, opened 2016-11-10
Update spec readme with the mongodb information

shivhg:
I ran into few issues while getting the specs pass in local, mainly about the missing mongoid information. I updated the Spec Readme(https://github.com/shivhg/cancancan/blob/develop/spec/README.rdoc).


coorasse:
This is to me a Operating System configuration. You would have also to write to have ruby installed or Postgres.

PR #363, opened 2016-11-17
Make accessible_by support multiple rules referencing the same table with different rules

### LINKED ISSUE #337: accessible_by not returning the correct results when multiple rules are present
I have a complex data structure with a large number of abilities. I have recently added two new abilities to add permissions for two new scenarios.

Model files:
```
class Asset < ActiveRecord::Base
  has_and_belongs_to_many :clients
  has_and_belongs_to_many :applications
end
```
```
class Client < ActiveRecord::Base
  belongs_to :firm
  belongs_to :owner, class_name: 'User'

  has_and_belongs_to_many :assets
  has_and_belongs_to_many :borrower_applications, class_name: 'Application', join_table: :borrowers
  has_and_belongs_to_many :guarantor_applications, class_name: 'Application', join_table: :guarantors
end
```
```
class Application < ActiveRecord::Base
  has_and_belongs_to_many :assets
  has_and_belongs_to_many :borrowers, class_name: 'Client', join_table: :borrowers
  has_and_belongs_to_many :guarantors, class_name: 'Client', join_table: :guarantors
end
```

So my assests can be linked to either a Client or an Application, and I want my permissions to be driven by `client.firm_id`, so a user in the right firm should have permissions on all Assets related to the `firms -> clients`, or the `firms -> clients -> applications` where the client is either a borrower or a guarantor.

So I have set up abilities like so:
```
firm_id = user.current_firm.try(:id)
can :manage, Asset, loan_applications: { guarantors: { firm_id: firm_id } }
can :manage, Asset, loan_applications: { borrowers: { firm_id: firm_id } }
can :manage, Asset, clients: { firm_id: firm_id }
```

So to test this I set up a client as a borrower on an application, and linked an asset to both the client and the application (as a borrower, not a guarantor)

Results:
```
firm_id = User.last.current_firm.id

Asset.last.clients.first.firm_id == firm_id #=> true
Asset.last.loan_applications.first.borrowers.first.firm_id == firm_id #=> true
Asset.last.loan_applications.first.guarantors.first.try(:firm_id) == firm_id #=> false

a = Ability.new(User.last)

a.can?(:read, Asset.last) #=> true
Asset.accessible_by(a, :read) #=> #<ActiveRecord::Relation []>

a.can?(:destroy, Asset.last) #=> true
Asset.accessible_by(a, :destroy) #=> #<ActiveRecord::Relation []>
```
Manually testing each of the rules suggest that two pass and one fails, but according to the ability precedence wiki page `Adding can rules do not override prior rules, but instead are logically or'ed.` so I would expect the user to have permissions on the asset.

While it is very possible that I am doing something wrong I dont know why I am getting different results from the `can?` and `accessible_by` methods.

For the record it works when I comment out the first two abilities.

mtrudel:
Fixes #337. This change refactors `ActiveRecordAdapter`, making it possible for two rules within the same [class, action] tuple to refer to the same table with different (and previously incompatible) conditions, and have the resulting relation contain the expected conditions. 

Essentially, each rule is now built out as a separate relation, with its own 'private' namespace for joins and conditions. The resulting relations are then combined using SQL's `UNION` and `EXCEPT` clauses into a master relation. Previously, all of the rules' conditions and joins were combined into  master condition and join expressions which were then expressed within a single relation, making it impossible for two separate rules to contain distinct sets of conditions against the same external table.

### DIFF
--- a/lib/cancan/model_adapters/active_record_3_adapter.rb
+++ b/lib/cancan/model_adapters/active_record_3_adapter.rb
@@ -5,12 +5,6 @@ module CanCan
       def self.for_class?(model_class)
         model_class <= ActiveRecord::Base
       end
-
-      private
-
-      def build_relation(*where_conditions)
-        @model_class.where(*where_conditions).includes(joins)
-      end
     end
   end
 end
--- a/lib/cancan/model_adapters/active_record_4_adapter.rb
+++ b/lib/cancan/model_adapters/active_record_4_adapter.rb
@@ -8,16 +8,6 @@ module CanCan
 
       private
 
-      # As of rails 4, `includes()` no longer causes active record to
-      # look inside the where clause to decide to outer join tables
-      # you're using in the where. Instead, `references()` is required
-      # in addition to `includes()` to force the outer join.
-      def build_relation(*where_conditions)
-        relation = @model_class.where(*where_conditions)
-        relation = relation.includes(joins).references(joins) if joins.present?
-        relation
-      end
-
       def self.override_condition_matching?(subject, name, value)
         # ActiveRecord introduced enums in version 4.1.
         (ActiveRecord::VERSION::MAJOR > 4 || ActiveRecord::VERSION::MINOR >= 1) &&
@@ -33,35 +23,6 @@ module CanCan
         value.is_a?(Enumerable) ? 
           (value.include? attribute) :
           attribute == value
-      end
-
-      # Rails 4.2 deprecates `sanitize_sql_hash_for_conditions`
-      def sanitize_sql(conditions)
-        if ActiveRecord::VERSION::MAJOR > 4 && Hash === conditions
-          table = @model_class.send(:arel_table)
-          table_metadata = ActiveRecord::TableMetadata.new(@model_class, table)
-          predicate_builder = ActiveRecord::PredicateBuilder.new(table_metadata)
-
-          conditions = predicate_builder.resolve_column_aliases(conditions)
-          conditions = @model_class.send(:expand_hash_conditions_for_aggregates, conditions)
-
-          conditions.stringify_keys!
-
-          predicate_builder.build_from_hash(conditions).map { |b|
-            @model_class.send(:connection).visitor.compile b
-          }.join(' AND ')
-        elsif ActiveRecord::VERSION::MINOR >= 2 && Hash === conditions
-          table = Arel::Table.new(@model_class.send(:table_name))
-
-          conditions = ActiveRecord::PredicateBuilder.resolve_column_aliases @model_class, conditions
-          conditions = @model_class.send(:expand_hash_conditions_for_aggregates, conditions)
-
-          ActiveRecord::PredicateBuilder.build_from_hash(@model_class, conditions, table).map { |b|
-            @model_class.send(:connection).visitor.compile b
-          }.join(' AND ')
-        else
-          @model_class.send(:sanitize_sql, conditions)
-        end
       end
     end
   end
--- a/spec/cancan/model_adapters/active_record_adapter_spec.rb
+++ b/spec/cancan/model_adapters/active_record_adapter_spec.rb
@@ -232,37 +232,37 @@ if defined? CanCan::ModelAdapters::ActiveRecordAdapter
       expect(lambda { @ability.can? :read, Article.new }).to raise_error(CanCan::Error)
     end
 
-    it "has false conditions if no abilities match" do
-      expect(@ability.model_adapter(Article, :read).conditions).to eq("'t'='f'")
+    it "has non-matching SQL if no abilities match" do
+      expect(@ability.model_adapter(Article, :read).database_records.to_sql.strip.squeeze(' ')).to eq(%q(SELECT "articles".* FROM "articles" WHERE (1 == 0)))
     end
 
-    it "returns false conditions for cannot clause" do
+    it "returns non-mathcing SQL if only clause is a single cannot clause" do
       @ability.cannot :read, Article
-      expect(@ability.model_adapter(Article, :read).conditions).to eq("'t'='f'")
+      expect(@ability.model_adapter(Article, :read).database_records.to_sql.strip.squeeze(' ')).to eq(%q(SELECT "articles".* FROM "articles" WHERE (1 == 0)))
     end
 
     it "returns SQL for single `can` definition in front of default `cannot` condition" do
       @ability.cannot :read, Article
       @ability.can :read, Article, :published => false, :secret => true
-      expect(@ability.model_adapter(Article, :read).conditions).to orderlessly_match(%Q["#{@article_table}"."published" = 'f' AND "#{@article_table}"."secret" = 't'])
+      expect(@ability.model_adapter(Article, :read).database_records.to_sql.strip.squeeze(' ').gsub(' )', ')')).to eq(%q(SELECT DISTINCT articles.* FROM (SELECT "articles".* FROM "articles" WHERE "articles"."published" = 'f' AND "articles"."secret" = 't' EXCEPT SELECT "articles".* FROM "articles") AS articles))
     end
 
-    it "returns true condition for single `can` definition in front of default `can` condition" do
+    it "returns SQL for single `can` definition in front of default `can` condition" do
       @ability.can :read, Article
       @ability.can :read, Article, :published => false, :secret => true
-      expect(@ability.model_adapter(Article, :read).conditions).to eq("'t'='t'")
+      expect(@ability.model_adapter(Article, :read).database_records.to_sql.strip.squeeze(' ')).to eq(%q(SELECT DISTINCT articles.* FROM (SELECT "articles".* FROM "articles" UNION SELECT "articles".* FROM "articles" WHERE "articles"."published" = 'f' AND "articles"."secret" = 't') AS articles))
     end
 
-    it "returns `false condition` for single `cannot` definition in front of default `cannot` condition" do
+    it "returns non-matching SQL for single `cannot` definition in front of default `cannot` condition" do
       @ability.cannot :read, Article
       @ability.cannot :read, Article, :published => false, :secret => true
-      expect(@ability.model_adapter(Article, :read).conditions).to eq("'t'='f'")
+      expect(@ability.model_adapter(Article, :read).database_records.to_sql.strip.squeeze(' ')).to eq(%q(SELECT "articles".* FROM "articles" WHERE (1 == 0)))
     end
 
-    it "returns `not (sql)` for single `cannot` definition in front of default `can` condition" do
+    it "returns SQL for single `cannot` definition in front of default `can` condition" do
       @ability.can :read, Article
       @ability.cannot :read, Article, :published => false, :secret => true
-      expect(@ability.model_adapter(Article, :read).conditions).to orderlessly_match(%Q["not (#{@article_table}"."published" = 'f' AND "#{@article_table}"."secret" = 't')])
+      expect(@ability.model_adapter(Article, :read).database_records.to_sql.strip.squeeze(' ')).to eq(%q(SELECT DISTINCT articles.* FROM (SELECT "articles".* FROM "articles" EXCEPT SELECT "articles".* FROM "articles" WHERE "articles"."published" = 'f' AND "articles"."secret" = 't') AS articles))
     end
 
     it "returns appropriate sql conditions in complex case" do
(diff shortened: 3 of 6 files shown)


truelson:
Would like to see this make it in. Thank you!

spec/cancan/model_adapters/active_record_adapter_spec.rb
```diff
@@ -232,96 +232,74 @@ class User < ActiveRecord::Base
...
     end
 
-    it "has false conditions if no abilities match" do
-      expect(@ability.model_adapter(Article, :read).conditions).to eq("'t'='f'")
+    it "has non-matching SQL if no abilities match" do
+      expect(@ability.model_adapter(Article, :read).database_records.to_sql.strip.squeeze(' ')).to eq(%q(SELECT "articles".* FROM "articles" WHERE (1 == 0)))
     end
 
-    it "returns false conditions for cannot clause" do
+    it "returns non-mathcing SQL if only clause is a single cannot clause" do
```
zachweed: Minor, but you have a typo in "matching".

mtrudel: Fixed...

FWMatt:
+1 to get this merged

zachweed:
@mtrudel let me know if you want me to jump in to get those conflicts resolved for ya

mtrudel:
@zachweed that would be awesome. Looks like there's a lot of upstream change that I haven't kept track of; probably easier for someone who knows upstream to review this. LMK if you need help / clarity!

zachweed:
sounds good. much appreciated!

mtrudel:
@zachweed your sub-PR is merged here; this is ready to go from my perspective.

zachweed:
thanks @mtrudel!

zachweed:
@mtrudel I'm fine to resolve conflicts again if we can get this merged in shortly afterwards?

mtrudel:
Conflict resolved - can someone upstream please take a look at this and hopefully merge?

coorasse:
can you please rebase on latest develop? thanks

lib/cancan/model_adapters/active_record_3_adapter.rb
```diff
@@ -5,12 +5,6 @@ class ActiveRecord3Adapter < AbstractAdapter
       def self.for_class?(model_class)
         model_class <= ActiveRecord::Base
       end
-
```
coorasse: this adapter has been removed

.rubocop_todo.yml
```diff
@@ -1,42 +1,65 @@
 # This configuration was generated by
 # `rubocop --auto-gen-config`
-# on 2016-12-14 08:11:19 +0100 using RuboCop version 0.45.0.
+# on 2017-03-12 12:11:26 -0400 using RuboCop version 0.47.1.
```
coorasse: rebase on latest develop to check against latest rubocop version. I would expect no changes to happen in this TODO except rare cases

lib/cancan/model_adapters/active_record_4_adapter.rb
```diff
@@ -6,14 +6,14 @@ def self.for_class?(model_class)
         model_class <= ActiveRecord::Base
       end
 
-      # TODO: this should be private
+      private
```
coorasse: this does not have any effect. rubocop also should tell

lib/cancan/model_adapters/active_record_4_adapter.rb
```diff
@@ -26,56 +26,6 @@ def self.matches_condition?(subject, name, value)
...
-
-        conditions.stringify_keys!
-
-        predicate_builder.build_from_hash(conditions).map do |b|
-          @model_class.send(:connection).visitor.compile b
-        end.join(' AND ')
-      end
     end
   end
 end
```
coorasse: this entire file could be actually be removed and merged into the active_record_adapter

spec/cancan/model_adapters/active_record_adapter_spec.rb
```diff
@@ -237,101 +237,84 @@ class User < ActiveRecord::Base
     end
 
     it 'has false conditions if no abilities match' do
-      expect(@ability.model_adapter(Article, :read).conditions).to eq("'t'='f'")
+      expect(@ability.model_adapter(Article, :read).database_records.to_sql.strip.squeeze(' '))
```
coorasse: the previous test should still be in place, right? with `expect(@ability.model_adapter(Article, :read).conditions).to eq("1 == 0")`

spec/cancan/model_adapters/active_record_adapter_spec.rb
```diff
@@ -237,101 +237,84 @@ class User < ActiveRecord::Base
...
+        .to eq(%q(SELECT "articles".* FROM "articles" WHERE (1 == 0)))
     end
 
     it 'returns false conditions for cannot clause' do
       @ability.cannot :read, Article
-      expect(@ability.model_adapter(Article, :read).conditions).to eq("'t'='f'")
+      expect(@ability.model_adapter(Article, :read).database_records.to_sql.strip.squeeze(' '))
```
coorasse: extract `.to_sql.strip.squeeze(' ')` into a method `to_sql(database_records)` and reuse it

spec/cancan/model_adapters/active_record_adapter_spec.rb
```diff
@@ -237,101 +237,84 @@ class User < ActiveRecord::Base
...
+        .to eq(%q(SELECT "articles".* FROM "articles" WHERE (1 == 0)))
     end
 
     it 'returns SQL for single `can` definition in front of default `cannot` condition' do
       @ability.cannot :read, Article
       @ability.can :read, Article, published: false, secret: true
-      expect(@ability.model_adapter(Article, :read).conditions)
-        .to orderlessly_match(%("#{@article_table}"."published" = 'f' AND "#{@article_table}"."secret" = 't'))
+      expect(@ability.model_adapter(Article, :read).database_records.to_sql.strip.squeeze(' ').gsub(' )', ')'))
```
coorasse: is `.gsub(' )', ')')` really necessary?

spec/cancan/rule_spec.rb
```diff
@@ -37,16 +37,4 @@
     rule = CanCan::Rule.new(true, :read, Integer, nil, nil)
     expect(rule.associations_hash).to eq({})
   end
-
-  it 'is not mergeable if conditions are not simple hashes' do
```
coorasse: why these tests got removed?

spec/cancan/model_adapters/active_record_adapter_spec.rb
```diff
@@ -237,101 +237,84 @@ class User < ActiveRecord::Base
...
       @ability.can :read, Article, ['secret=?', false]
       adapter = @ability.model_adapter(Article, :read)
       2.times do
-        expect(adapter.conditions).to eq(%[(secret='f') OR ("#{@article_table}"."published" = 't')])
+        expect(adapter.database_records.to_sql.strip.squeeze(' '))
+          .to eq(%q(SELECT DISTINCT articles.* FROM (SELECT "articles".* FROM "articles" WHERE "articles"."published" = 't' UNION SELECT "articles".* FROM "articles" WHERE (secret='f')) AS articles))
       end
     end
 
-    it 'has nil joins if no rules' do
```
coorasse: why these tests got removed?

spec/cancan/model_adapters/active_record_adapter_spec.rb
```diff
@@ -237,101 +237,84 @@ class User < ActiveRecord::Base
...
+        .to eq(%q(SELECT DISTINCT articles.* FROM (SELECT "articles".* FROM "articles" WHERE "articles"."id" = 1) AS articles))
+      expect(@ability.model_adapter(Article, :read).database_records.to_sql.strip.squeeze(' '))
+        .to eq(%q(SELECT DISTINCT articles.* FROM (SELECT "articles".* FROM "articles" UNION SELECT "articles".* FROM "articles" WHERE "articles"."id" = 1) AS articles))
+    end
+
+    it 'returns appropriate sql conditions in cases where two rules specify different conditions on the same remote table, accessed via distinct joins' do
+      @ability.can :read, Article, user: { id: 1 }
+      @ability.can :read, Article, { mentioned_users: { id: 2 } }
+      expect(@ability.model_adapter(Article, :read).database_records.to_sql.strip.squeeze(' '))
+        .to eq(%q(SELECT DISTINCT articles.* FROM (SELECT "articles".* FROM "articles" INNER JOIN "users" ON "users"."id" = "articles"."user_id" WHERE "users"."id" = 1 UNION SELECT "articles".* FROM "articles" INNER JOIN "legacy_mentions" ON "legacy_mentions"."article_id" = "articles"."id" INNER JOIN "users" ON "users"."id" = "legacy_mentions"."user_id" WHERE "users"."id" = 2) AS articles))
```
coorasse: why does it use a `union`? shouldn't perform more joins?

lib/cancan/model_adapters/active_record_3_adapter.rb
mtrudel: Noted. I'll remove this change

lib/cancan/model_adapters/active_record_4_adapter.rb
mtrudel: Per previous comment, I'll do so.

spec/cancan/model_adapters/active_record_adapter_spec.rb
mtrudel: No, it shouldn't still be in place - the entire underlying mechanism for generating queries has changed, and in such a way that the previous placeholder 't=f' query isn't present any more.

spec/cancan/model_adapters/active_record_adapter_spec.rb
mtrudel: Will do.

spec/cancan/model_adapters/active_record_adapter_spec.rb
mtrudel: IIRC this was because AR3 and AR4 adapters produced queries that differed only in whitespace, so I normalized them to a standard form. With the consolidation of the AR drivers I'll revisit this and remove if possible

spec/cancan/model_adapters/active_record_adapter_spec.rb
mtrudel: Nope - UNIONs are at the heart of what this PR enables. Recall from the PR notes that previously cancancan would split off the logic that generates joins from the condition logic quite early in the query generation process, making it impossible to produce queries on a per-join basis. The heart of this PR is inverting that behaviour and instead producing subqueries for each ability with the requisite joins and conditions being self-contained. These subqueries are then joined via SQL UNION statements to produce the desired net effect

spec/cancan/model_adapters/active_record_adapter_spec.rb
mtrudel: See previous - this test is obsolete under the new query model

mtrudel:
@coorasse Changes are rebased in. I ended up manually reapplying the changes (and cleaning up the branch quite a bit in the process), so it may be worth starting review from scratch. All of your concerns were addressed, but since the branches have diverged so much Github didn't catch all of the changes.

mvi-michael:
Thanks @mtrudel! +1 for this.

This has been solving my problems but I have a question about recent commits: https://github.com/CanCanCommunity/cancancan/commit/05b0f058aaf0126c5d00fdf9f45b8af1b861cd4f#diff-9461ac1737603a6d4ec970b209028ee7R33

Should this
`@model_class.where('1 == 0')` 
be changed to this
`@model_class.where('1 = 0')` 
or even 
`@model_class.where(false)`

Also the use of `DISTINCT`s a few lines down are causing me issues when then trying to implement an `ORDER BY`. For example: `ORDER BY expressions must appear in select list`. 
Is `DISTINCT` necessary if the query is built with `INNER JOIN`?

mtrudel:
@mvi-michael `.where(false)` (confusingly, IMO) returns a query with an empty `WHERE` clause, the exact opposite of what we want. Point taken about single equals, however. While it works on postgres, the single equals is more correct. I'll push a change to that momentarily.

Unsure about the specifics of your question regarding `DISTINCT` - can you provide a specific example?

mtrudel:
Also a friendly reminder to @coorasse or other maintainers that the previously asked for changes have been made and this branch is ready for review.

mvi-michael:
@mtrudel I think I wrongly assumed the `DISTINCT` was the sole issue when it was a combination of factors. My query effectively involved calling an `ORDER BY (LOWER(table.column_name))` on the results of `accessible_by`, which meant I had an *expression* not included in the select list - hence my error. 

I've sorted it now and you can disregard the `DISTINCT` issue. My bad. 
Thanks for the "single equals" change.

cbisnett:
So I spent some time trying to get a solution working for this same problem. I found a patch to a similar issue in an old CanCan issue (ryanb/cancan#213). I was able to apply the patch with a few modifications and it successfully solved the issue.

What's the likelihood that this will get changes and get merged? I'm in a situation where I need this functionality and I was looking to submit a PR when I found this one. Would it be better for me to try to get this one across the line and merged or should I submit a different PR with my patch with fewer changes?

mtrudel:
@cbisnett I've rebased this onto significant upstream changes three times now in the hopes of getting it merged, with no action from the maintainers. That, plus the fact that I'm no longer on the project where this was originally used means that I'm not super keen on working on this any further (at least not without upstream committers expressing some support of this eventually landing). 

That having been said, it's still a valid improvement in my eyes (fixes the original issue, plus it's cleaner internally). If you're keen to champion this, I'm happy to support.

mtrudel:
@cbisnett to your second point, I think continuing on this PR is worth it over and above the actual issue fix, largely due to how much easier it is to test & reason about internally.

cbisnett:
I've pushed our branch with the updated patch to [our fork](https://github.com/huntresslabs/cancancan/commit/db406074262bb479a9d1bb3b71710aca7688276a). The main difference between this patch and the one I've updated, is that the one I've updated makes use of Arel to build the unions (rather than `to_sql`) and updates the `bind_values`. Also, it makes fewer changes to the tests and internal workings. Currently the patch is applied to the v1.13.1 tag but could be ported to the `develop` branch with some work.

I'll leave it up to the maintainers to decide if they are interested in this work and how they want to integrate it.

coorasse:
I will not have time to look into it properly in the next two weeks. 
As soon as I'll get back to work on cancancan I'll test this version in a couple of very complex projects to see if everything still works as expected. Thank you very much for your effort.

coorasse:
I have some major concerns about performances. As far as I see from this implementation each condition would be extracted into a separate query and conditions not connected by an OR but by UNION or EXCEPT. Am I right?

coorasse:
Could you please provide some benchmarks of these changes?

cbisnett:
We've been using a slight variation on this at Huntress Labs in production for several months. We use this ability to power our permissions model. I would like to get it upstreamed at somepoint and will make a new pull request, but before I do, what type of benchmarks would you like to see?

coorasse:
Looks like some more complicated conditions would benefit a lot from this PR.
At the moment, `cancancan` performs a `LEFT OUTER JOIN` in order to be able to combine multiple conditions. This can cause, in large databases or more complex conditions a very slow query.
By using your version (with UNION) in my own projects I noticed a great benefit in terms of performance.
Nevertheless, this PR may have a huge impact on existing projects, therefore I am delaying its approval.

coorasse:
I identified in #514 some tests that would not pass with this implementation.
I plan to merge this within version 3.0.0, and I already proposed some changes that would fix these issues in branch `feature/union`. Next steps will be:
* point those changes to version 3.0.0
* write some benchmarks to prove the actual benefit of this implementation
* add tests where we use other libraries and methods together with cancancan. some examples:
** used in combination with `includes`
** used in combination with scopes
** used in combination with `pg_search`
** ...

cbisnett:
@coorasse Are you planning on making these changes yourself or are you looking for someone else to make the changes?

This patch converts all the relations to the SQL string and joins them using the string `' UNION '`. Is that what you're looking for? In a separate implementation, I've done the work to build these using ActiveRecord and Arel. Is that something that should be considered?

coorasse:
I am planning to reuse your code and open a pull request myself towards version 3.0.0. I am not able to release this before 3.0.0 because that branch contains already improvements to the "left joins method". If the benchmarks confirm me that this is actually a better implementation as I think, it will replace the LEFT JOINS completely in version 3.0.0. This means a further delay for the release of 3.0.0, that I planned to release during this week, but I prefer it like that.
Please check [here](https://github.com/CanCanCommunity/cancancan/pull/516/files#diff-a247715d41249fbfb6d418e6e18c939dR115) the regressions tests I added, because the "UNION version" would break them, and [here](https://github.com/CanCanCommunity/cancancan/blob/feature/union/lib/cancan/model_adapters/active_record_adapter.rb) the version I wrote yesterday starting from your code that fixes the issue (WIP). I will keep this PR open until another one is ready. Thank you!

cbisnett:
I'm not terribly surprised at those performance benchmarks. I have yet to dig into the benchmark code to double check that nothing was missed that would change the performance (missing indices, sub-optimal queries, etc.) but I think this is probably pretty standard for SQL unions.

I think what's missing here is some code to detect when a `LEFT JOIN` should be used and when a `UNION` should be used. I've got some ideas for detecting these situations and then I think it might be possible to make some changes to allow this to be explicitly requested.

coorasse:
Those benchmarks surprised me because in two of my apps I actually noticed a reduced execution time of the SQL queries, so either the benchmarks are not well structured, or the databases of my apps aren't. Apart from that I, now I have in mind two things as well:
1. Why don't we use union only in case of "conflicts" when the same table is joined twice and therefore we are not currently able to return the correct result
2. Why are considering UNION at all? Why not fix the issue by using Arel and keeping the left joins?

cbisnett:
It's unlikely that using SQL `UNION` will have better performance than `LEFT JOIN` in most cases, because each query has to be separately executed and the results aggregated afterwards. I think in almost all cases `LEFT JOIN` should be used except in situations where we detect that multiple rules apply and their joins don't match. It's likely these joins can't be combined into a query that produces the correct results.

I've written code to handle the situation where we need to union multiple ActiveRecord relations for both Rails 5.1 and 5.2. Depending on how far back we want to handle, I've also got code that works against 4.2 (tested on 4.2.7 and 4.2.10). This week I'll turn what I've got into a pull request with some code to detect the situations that need unions rather than left joins and we can see how well it works.

coorasse:
Hey @cbisnett, I'd still be interested in seeing how you identify these situations, in order to bring on this topic and support this case as well

PR #370, opened 2016-12-11
drop support for InheritedResource

coorasse:
replaces https://github.com/CanCanCommunity/cancancan/pull/266/files since InheritedResource is not supported anymore. see: https://github.com/josevalim/inherited_resources#deprecation-notice


teohm:
@coorasse @schmijos Hi, I noticed CanCanCan 2 has now dropped support for InheritedResource. We are still using InheritedResource together with CanCanCan in production. To upgrade to CanCanCan 2 and Rails 5, is there an available upgrade path for us?

coorasse:
no,  sorry. we don't support it anymore. I suggest you to monkey patch cancancan in your project, even if this can tie you to old, unsupported, libraries even more.

varyonic:
Just as an FYI inherited_resources was transferred to the [activeadmin](https://github.com/activeadmin/inherited_resources) organization for continuing maintenance about a month after this PR was merged.  New features are discouraged but Rails 5.2 is supported.

TylerRick:
Hi @teohm, did you figure out a good upgrade path? I'm in a similar situation, upgrading to latest Rails and CanCanCan, only to discover that CanCanCan no longer works with InheritedResource (even though InheritedResource is still maintained and used). 

I guess I'll try copying the code that was in earlier version of CanCanCan and see if it still works... :disappointed:

TylerRick:
Fortunately there wasn't a very big integration layer...
```
module CanCan
  module ControllerAdditions
    module ClassMethods
      def cancan_resource_class
        if ancestors.map(&:to_s).include? "InheritedResources::Actions"
          InheritedResource
        else
          ControllerResource
        end
      end
    end
  end
end

module CanCan
  class InheritedResource < ControllerResource
    def load_resource_instance
      if parent?
        @controller.send :association_chain
        @controller.instance_variable_get("@#{instance_name}")
      elsif new_actions.include? @params[:action].to_sym
        resource = @controller.send :build_resource
        assign_attributes(resource)
      else
        @controller.send :resource
      end
    end

    def resource_base
      @controller.send :end_of_association_chain
    end
  end
end
```

coorasse:
A new gem is welcome, I don't feel ike adding it back to the source code. It has been eliminated because the project was abandoned. Now the project is maintained again but they also suggest not to use it so...

TylerRick:
I understand... :) A new gem is a good idea.

TylerRick:
Here is one: https://rubygems.org/gems/cancan-inherited_resources

PR #378, opened 2016-12-20
Fix rubocop Issues: Lint/EndAlignment

matthiasfehr:


lib/cancan/controller_resource.rb
```diff
@@ -229,10 +229,13 @@ def name
     def resource_params
       if parameters_require_sanitizing? && params_method.present?
         return case params_method
-          when Symbol then @controller.send(params_method)
-          when String then @controller.instance_eval(params_method)
-          when Proc then params_method.call(@controller)
-        end
+                 when Symbol then
```
coorasse: I think the `then` keyword is not necessary. can you remove it?

PR #388, opened 2017-01-17
Add permissions_hash to ability

lchanmann:
To extract Hash of conditions from Ability.rb

### DIFF
--- a/lib/cancan/ability.rb
+++ b/lib/cancan/ability.rb
@@ -278,6 +278,21 @@ module CanCan
       permissions_list
     end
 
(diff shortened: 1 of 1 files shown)


lchanmann:
@coorasse it was an un-finished experiment and I didn't mean to submit this PR. My apology for that :)

BTW, I was trying to export cancan ability definitions to other format (eg. JSON) to be used on the client side such that I don't have to duplicate the rules defined in Ability.rb in my single-page app.

Do you know is there a build-in way or any WIP to accomplish that? Thanks!

PR #398, opened 2017-02-28
MongoidAdapter: Improves querying for finding relation_ids by replacing only->map with distinct

shaunakpp:
`distinct(:_id)` gives a significant performance boost over `only(:id).map(:id)`
I've also added a Benchmark for this.
```
1_000.times { FactoryGirl.create(:user) }
```
```
Benchmark.bmbm do |benchmark|
  benchmark.report("with_only_and_map") do
    User.only(:id).map(&:id)
  end
  benchmark.report("with_distinct") do
    User.distinct(:_id)
  end
end

```

```
user                system     total      real
with_only_and_map   0.020000   0.000000   0.020000 (  0.021605)
with_distinct       0.000000   0.000000   0.000000 (  0.012682)
```

---
This change is [Reviewable](https://reviewable.io/reviews/cancancommunity/cancancan/398)


coorasse:
Nice! 👍

PR #409, opened 2017-05-05
Fix issue when you pass a NULL param to the controller in Rails 5

### LINKED ISSUE #408: empty string in Rails 5 parameters will cause load_resource to raise error
I'm seeing `load_resource` raising errors when the value for the specified key is `nil` on the request.

### Example
Controller
```ruby
class UserController < ApplicationController
  load_resource :user, only: [:create, :destroy]

  def create; end
  def destroy; end
end
```
Request
```
POST :create, params: { user_id: nil }
```

In Rails 5, param values passed to the controller would all be interpreted as strings. Therefore, if `user_id` in the request above will look like `{ user_id: '' }` inside the controller. This will cause `load_resource` to raise an error because CanCan will try to do a `User.find('_id')`

https://github.com/CanCanCommunity/cancancan/blob/v1.15.0/lib/cancan/controller_resource.rb#L136-L146

darwin67:
resolves #408

### DIFF
--- a/lib/cancan/controller_resource.rb
+++ b/lib/cancan/controller_resource.rb
@@ -128,7 +128,7 @@ module CanCan
     end
 
     def id_param
-      @params[id_param_key].to_s if @params[id_param_key]
+      @params[id_param_key].to_s if @params[id_param_key].present?
     end
 
     def id_param_key
--- a/spec/cancan/controller_resource_spec.rb
+++ b/spec/cancan/controller_resource_spec.rb
@@ -412,6 +412,13 @@ describe CanCan::ControllerResource do
       expect(controller.instance_variable_get(:@model)).to be_nil
     end
 
+    it 'does not try to load resource for other action if params[:id] is blank' do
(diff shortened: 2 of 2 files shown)


lib/cancan/controller_resource.rb
```diff
@@ -138,7 +138,7 @@ def parent_authorization_action
     end
 
     def id_param
-      @params[id_param_key].to_s if @params[id_param_key]
+      @params[id_param_key].to_s if @params[id_param_key].present?
```
coorasse: can you please add a test which fails before your fix and that is fixed now? I also change the base to `feature/v2` branch so we can release this with next major release. Thanks!

darwin67:
@coorasse added the test for the change

darwin67:
i'm pretty sure the test failure have nothing to do with my changes...

darwin67:
looks like there's an invalid test configuration for Rails 5.
Rails 5 doesn't work below Ruby `2.2.2`

PR #416, opened 2017-05-31
Fix i18n lookup for unauthorized message

### LINKED ISSUE #415: I18n::ArgumentError with i18n v0.8.2+
### Steps to reproduce

https://gist.github.com/cbeer/140cb717506e97181644c59faaef92b2

Using i18n v0.8.2+, calling `unauthorized_message` causes an exception.

Upstream, they introduced a new check in 0.8.2: https://github.com/svenfuchs/i18n/pull/236.

Cancancan is explicitly passing a `nil` key: https://github.com/CanCanCommunity/cancancan/blob/2d21842d4415913cb9d15d71bfa3165087be29c4/lib/cancan/ability.rb#L225.

### Expected behavior

The unauthorized message should be provided.

### Actual behavior

An exception is thrown:

```
I18n::ArgumentError: I18n::ArgumentError
    /Users/cabeer/.gem/ruby/2.4.1/gems/i18n-0.8.3/lib/i18n.rb:152:in `translate'
    /Users/cabeer/.gem/ruby/2.4.1/gems/cancancan-2.0.0/lib/cancan/ability.rb:226:in `unauthorized_message'
```

### System configuration
**Rails version**: 5.0.2

**Ruby version**: 2.3.4

**CanCanCan version** 1.17.0

clemens:
Fixes #415 without relying on specific i18n version.

Inspired by https://github.com/plataformatec/simple_form/blob/master/lib/simple_form/inputs/base.rb#L173-L189.

### DIFF
--- a/lib/cancan/ability.rb
+++ b/lib/cancan/ability.rb
@@ -223,7 +223,7 @@ module CanCan
       keys = unauthorized_message_keys(action, subject)
       variables = { action: action.to_s }
       variables[:subject] = (subject.class == Class ? subject : subject.class).to_s.underscore.humanize.downcase
-      message = I18n.translate(nil, variables.merge(scope: :unauthorized, default: keys + ['']))
+      message = I18n.translate(keys.shift, variables.merge(scope: :unauthorized, default: keys + ['']))
       message.blank? ? nil : message
     end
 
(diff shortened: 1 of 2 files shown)


coorasse:
Thank you @clemens, I think we need a test here to proove that the issue is really solved, so a test that would actually fail without your changes and passes with them. Thanks for your contribution! Really appreciated.

clemens:
What do you have in mind for that test? A mock that expects a call with a certain key? Otherwise we'd have to test against a specific version of i18n – which seems questionable.

PR #427, opened 2017-06-27
Raise error when ability definition includes regex

jaredbeck:
Previously, regexes silently failed to do what you'd expect. Example:

    # consider this rule ..
    can :eat, Banana, species: /Musa/
    # and this banana ..
    Banana.new(species: 'Musa acuminata')
    # Rule#condition_match? will attempt the following comparison
    'Musa acuminata' == /Musa/

Clearly, the `==` operator is not effective here.

We could also raise this error in other places, but if we do, I think we
should still raise it here, in `condition_match?`.

### DIFF
--- a/CHANGELOG.rdoc
+++ b/CHANGELOG.rdoc
@@ -1,5 +1,7 @@
 Unreleased
 
+* Raise error when an ability definition includes a regex. Previously, regexes
+  silently failed to do what you'd expect.
 * Adds support for Rails Api applications (coorasse & ajgon)
 * Controller subclasses inherit skip_load_resource from superclass (jpmckinney)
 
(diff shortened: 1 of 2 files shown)


coorasse:
can you please provide a gist following the issue template? This would help me a lot. Thanks.

jaredbeck:
> can you please provide a gist following the issue template? This would help me a lot. Thanks.

No problem, here you are: https://gist.github.com/jaredbeck/da27e8d2135ca85b3f5db4067f29d986

We use a similar template for PaperTrail, aren't they great?

coorasse:
what if, instead, we add support for them?

jaredbeck:
Hi, I opened this issue over a year ago. It's not something I want to work on anymore, so I'll close it.

PR #439, opened 2017-08-29
Add ability to build new instances with `new_params` on `new` action

joelvh:
This enhancement provides the ability to specify the default attributes that are used when instantiating a new instance on the `new` action. Currently, `ControllerResource#resource_params` does not consider the `new` action as needing "sanitization", therefore it doesn't detect the `new_params` method on the controller if you've specified it.

This is a simple update to add the ability to specify defaults with `new_params` method on your controller.

### DIFF
--- a/lib/cancan/controller_resource.rb
+++ b/lib/cancan/controller_resource.rb
@@ -243,9 +243,21 @@ module CanCan
         when Proc then
           params_method.call(@controller)
         end
+      elsif build_params_method
+        @controller.send(build_params_method)
       else
         resource_params_by_namespaced_name
       end
+    end
+
+    def build_params_method
+      return unless new_actions.include?(@params[:action].to_sym)
+
+      method_name = "#{@params[:action]}_params".to_sym
+
+      return unless @controller.respond_to?(method_name, true)
+
+      method_name
     end
 
     def parameters_require_sanitizing?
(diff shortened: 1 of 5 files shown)


spec/cancan/controller_resource_spec.rb
```diff
@@ -185,6 +185,34 @@ class HiddenModel < ::Model; end
...
+        params[:action] = 'create'
+        allow(controller).to receive(:resource_params).and_return(resource: 'params')
+        allow(controller).to receive(:model_params).and_return(model: 'params')
+        allow(controller).to receive(:custom_params).and_return(custom: 'params')
+        allow(controller).to receive(:new_params).and_return(name: 'new')
+        allow(controller).to receive(:create_params).and_return(name: 'create')
+        resource = CanCan::ControllerResource.new(controller)
+        resource.load_resource
+        expect(resource.send(:build_params_method)).to eq :create_params
+        expect(resource.send(:resource_instance).name).to eq 'create'
```
coorasse: a bit too much duplicated code in this test. could you please refactor it? also the description of second `it` block is wrong.

lib/cancan/controller_resource.rb
```diff
@@ -243,11 +243,23 @@ def resource_params
...
         end
+      elsif build_params_method
+        @controller.send(build_params_method)
       else
         resource_params_by_namespaced_name
       end
     end
 
+    def build_params_method
+      return unless new_actions.include?(@params[:action].to_sym)
```
coorasse: this should not affect the create action but only new.

coorasse:
This change needs to be reconsidered with the changes included in v. 3.0

joelvh:
Hi @coorasse haven’t used this in a while - what’s new in 3.0?

PR #444, opened 2017-09-22
Allow strings as enum values

s-mage:
Active Record 5 improved work with enums. So now we can use strings or symbols instead of integers. For example:

```
# ActiveRecord4Adapter
@ability.can :read, Shape, color: Shape.colors[:green]

# ActiveRecord5Adapter
@ability.can :read, Shape, color: :green
```

`ActiveRecord5Adapter` is compatible with `ActiveRecord4Adapter` so you can use numbers as well.


Schwad:
@s-mage do you think these checks are failing for [ruby 2.0 here](https://travis-ci.org/CanCanCommunity/cancancan/jobs/278529288) because of your changes or has CanCanCan 2.0 dropped support for Ruby 2.0?

Schwad:
As a fan of `enums` and `Rails 5` I think this would be a handy addition!

s-mage:
@Schwad https://github.com/ruby/ruby/blob/ruby_2_3/NEWS I looked through the file and can't see anything incompatible in the diff, do you?

s-mage:
@Schwad merged develop there and it started fail in the same rubies where develop fails

s-mage:
hi @coorasse, it's been a while. I've added a patch for working with enums you described, now it's consistent (works both for keys and values).

coorasse:
scheduled for version 2.4.0. Thanks! 🙇

PR #457, opened 2017-11-20
Add Rails 5 API support in proper way

Eric-Guo:
Fix #418 introduce bug

[Application level before this PR failed result](https://travis-ci.org/ruby-china/homeland/builds/304744079) and [after result](https://github.com/ruby-china/homeland/pull/973)


coorasse:
Thank you!

Issue #413, opened 2017-05-26
Change in rules behaviour affects RailsAdmin root actions


dark-panda:
The change in c81cf44861e4f849af2fff92b9b9a17dac25796d to `Cancan::Rule#initialize` has introduced a small behavioural change that is affecting any authorization calls involving a `nil` subject. This was discovered using a custom RailsAdmin action that is root-level and thus has no real subject. In these cases, the calls to `#can?` have an action but just a `nil` subject.

The main difference in the behaviour can be seen in how the `@subjects` value is initialized when the `subject` argument is `nil`:

```ruby
@subjects = [subject].flatten
# => [nil]

@subjects = Array(subject)
# => []
```

Because the `@subjects` value is an empty array as opposed to an array with a single `nil` element, none of the rules that would normally match against it are matching.

Perhaps the following would suffice as an easy patch?

```ruby
@subjects = if subject.nil?
  [nil]
else
  Array(subject)
end
```

coorasse:
can you please provide a test that shows the expected behaviour and what actually happens? You can use this gist: https://gist.github.com/coorasse/3f00f536563249125a37e15a1652648c

I see that the subjects are now `[]` instead of `[nil]` but this is internal behaviour and shouldn't change anything for you. I need to see a more real-life case of what is happening. Thanks

dark-panda:
Sure, I could write out a test, but would the test be from the RailsAdmin perspective specifically? Basically the behaviour is exactly what you describe where subjects are `[]` versus `[nil]`. When creating custom RailsAdmin actions that return `true` for `root?`, there is no subject either as a model or as an instance, and the value that RailsAdmin passes is `nil`. Since there are no subjects to check on when the subjects value is `[]`, the check always returns `false`.

At any rate, I can write out the tests, although perhaps a backtrace with method arguments from RailsAdmin might also be useful as well? What it basically boils down to is that RailsAdmin is asking `can?(:root_action_name, nil)` for any root-level actions when checking for authorization, and as near as I can tell the empty subjects array causes those checks to always return `false` regardless of the abilities configuration.

coorasse:
I think this is a specific case for RailsAdmin. I don't see documented anywhere else the usage of `can?` without a subject. So, basically, RailsAdmin uses `can? :dashboard, nil`, am I right?

coorasse:
[...]

PR #460, opened 2017-11-20
Subject is required when action is provided

andrew-aladev:
#413


lib/cancan/rule.rb
```diff
@@ -17,6 +17,7 @@ def initialize(base_behavior, action, subject, conditions, block)
                                   "#{action} #{subject} ability. Use either one."
       raise Error, both_block_and_hash_error if conditions.is_a?(Hash) && block
       @match_all = action.nil? && subject.nil?
+      raise Error, 'Subject is required for action' if !action.nil? && subject.nil?
```
coorasse: nice. could you also print the action?

coorasse:
Please check your Travis build, is failing. thanks.

coorasse:
This is a different path to support rules without subject (see https://github.com/CanCanCommunity/cancancan/pull/417). 
We should decide onto something. I would prefer this way (do not support rules without subject) but I am open to proposals

coorasse:
this has been implemented already in 3.0.0. Thanks for your contribution!

Issue #456, opened 2017-11-20
Doesn't raise exception if resource is not found


camelmasa:
### Steps to reproduce

Use `load_and_authorize_resource` method on Rails app, like below:

```rb
class UsersController < ApplicationController
  load_and_authorize_resource find_by: :uuid

  def show
  end
end
```

And then you access to not exist user path. (e,g, http://localhost:3000/users/not_exist_user_uuid)

I think this line  https://github.com/CanCanCommunity/cancancan/commit/33df7eb51a85692e702dc0e2a7252471bfe271ef#diff-38607cbe9e3608bd14d9367c9cf91eefL110 changed behavior

The line will raise exception if resource is not found.

### Expected behavior

Raise exception if resource is not found

### Actual behavior

Doesn't raise exception if resource is not found

### System configuration
**Rails version**: 5.1.4

**Ruby version**: 2.4.2p198 (2017-09-14 revision 59899) [x86_64-darwin16]

**CanCanCan version** 2.1.1

camelmasa:
@coorasse Let me know if the change is not specification. I'd like to create PR 😃

lostapathy:
I can confirm that there is a behavior regression here between cancancan 2.1.0 and 2.1.1.  I'd go ahead and write a test and make that PR.

coorasse:
go for it. thanks for the catch

camelmasa:
@lostapathy Oh, I created PR for this. Thanks. 

@coorasse I created PR for fixing this. Could you review PR when you have time 🕐  https://github.com/CanCanCommunity/cancancan/pull/461

coorasse:
I know understand better the issue. Maybe we could replace `find_by` with `find_by!` so we keep raising an exception but we don't reintroduce dynamic finders. what do you think?

lostapathy:
+1 @coorasse Without testing it, that sounds like a good plan.

Should also keep the change to call .find_by! directly rather than invoking it via .send

camelmasa:
@coorasse @lostapathy Sounds good 👍 I'll update files.

camelmasa:
> Maybe we could replace find_by with find_by! so we keep raising an exception but we don't reintroduce dynamic finders.

@coorasse Did you mean we can change `find_resource_using_find_by` like below ❓ 

```rb
    def find_resource_using_find_by
      if resource_base.respond_to? "find_by!"
        resource_base.find_by!(@options[:find_by].to_sym => id_param)
      else
[...]

PR #461, opened 2017-11-20
Fix bug which is it doesn't raise exception if resource is not found

camelmasa:
Fixed https://github.com/CanCanCommunity/cancancan/issues/456


spec/cancan/controller_resource_spec.rb
```diff
@@ -501,10 +501,10 @@ class Model < ::Model; end
...
     it 'allows full find method to be passed into find_by option' do
       model = Model.new
-      allow(Model).to receive(:find_by_name).with('foo') { model }
+      allow(Model).to receive(:find_by_name!).with('foo') { model }
 
       params.merge!(action: 'show', id: 'foo')
-      resource = CanCan::ControllerResource.new(controller, find_by: :find_by_name)
+      resource = CanCan::ControllerResource.new(controller, find_by: :name)
       resource.load_resource
       expect(controller.instance_variable_get(:@model)).to eq(model)
```
camelmasa: This test case was not worked before. It was same as https://github.com/CanCanCommunity/cancancan/pull/461/files#diff-eba4e6277d1f300a2baa8a35cef7ef0fR492.

so I updated test case.

lib/cancan/controller_resource_finder.rb
```diff
@@ -13,7 +13,9 @@ def find_resource
     end
 
     def find_resource_using_find_by
-      if resource_base.respond_to? 'find_by'
+      if resource_base.respond_to? "find_by_#{@options[:find_by]}!"
+        resource_base.send("find_by_#{@options[:find_by]}!", id_param)
+      elsif resource_base.respond_to? 'find_by'
```
camelmasa: Revived it from here https://github.com/CanCanCommunity/cancancan/commit/33df7eb51a85692e702dc0e2a7252471bfe271ef#diff-38607cbe9e3608bd14d9367c9cf91eefL110

lib/cancan/controller_resource_finder.rb
```diff
@@ -13,7 +13,9 @@ def find_resource
...
         resource_base.send('find_by', @options[:find_by].to_sym => id_param)
```
lostapathy: I know it didn't start in this PR, but why are we calling .send() on this rather than going direct in this case, such as:

resource_base.find_by(@options[:find_by].to_sym => id_param)

camelmasa: Right, I think we can call `find_by` directly like that 👍  That would be great.

lib/cancan/controller_resource_finder.rb
```diff
@@ -13,8 +13,10 @@ def find_resource
     end
 
     def find_resource_using_find_by
-      if resource_base.respond_to? 'find_by'
-        resource_base.send('find_by', @options[:find_by].to_sym => id_param)
+      if resource_base.respond_to? "find_by_#{@options[:find_by]}!"
```
coorasse: support for dynamic finders has been dropped.

lib/cancan/controller_resource_finder.rb
```diff
@@ -13,8 +13,10 @@ def find_resource
...
+        resource_base.send("find_by_#{@options[:find_by]}!", id_param)
+      elsif resource_base.respond_to? 'find_by'
+        resource_base.find_by(@options[:find_by].to_sym => id_param)
```
coorasse: 👍

spec/cancan/controller_resource_spec.rb
coorasse: this is actually a bit misleading. could you restore the intention of the previous test? but maybe use `allow(Model).to receive(:search).with('foo') { model }` on line 504 and on line 507 you use `find_by: :search` which should reflect the previous behaviour.

spec/cancan/controller_resource_spec.rb.orig
```diff
@@ -1,708 +0,0 @@
-require 'spec_helper'
```
coorasse: how did this end up there? 😅

coorasse:
Please look at my comment: https://github.com/CanCanCommunity/cancancan/issues/456#issuecomment-346081557 I think I misinterpreted partially your PR while reviewing it

spec/cancan/controller_resource_spec.rb.orig
camelmasa: I found that on  https://github.com/CanCanCommunity/cancancan/pull/452/commits/1e6b5705d1a57b5e49faa1a2732291802ba72be9 😃

spec/cancan/controller_resource_spec.rb
camelmasa: Ah I see. I didn't notice that. Got it.

camelmasa:
@coorasse  Thanks for comments. I updated this PR. Could you review when you have time ❓ 

Also, I changed one test case on https://github.com/CanCanCommunity/cancancan/pull/461/commits/5c9d21df8c18e1befaa3699e72032cc04eafa841.
Is my understanding correct ❓

coorasse:
Following @terryyin comment (https://github.com/CanCanCommunity/cancancan/commit/8fc2a2dbefc542749650da6b957335e8075b34a3#commitcomment-25775734) we have a regression at the moment. I am restoring the old behaviour now through another PR and then we can look at a refactoring which would deprecate the existing behaviour and therefore fall into a major release

coorasse:
could you please check your PR and update it against the current develop branch? Thanks

terryyin:
The dynamic finders with ! is still supported in Rails 5 (http://edgeguides.rubyonrails.org/active_record_querying.html#dynamic-finders). Perhaps there's no need to change this to use "find_by!" now, because it still changes the current behaviour (people might have their dynamic finder override).

camelmasa:
@terryyin Thanks for advice 😃 

@coorasse I confirmed latest version 2.1.2 is released. I'll close it. Thanks for releasing.

terryyin:
Thanks!

Great work, BTW.

lostapathy:
Thanks @coorasse for taking care of this

PR #462, opened 2017-11-21
Add support to translate model_name

nyamadori:
This PR allows to translate a model's name in unauthorized message using ActiveSupport I18n.

### DIFF
--- a/lib/cancan/ability.rb
+++ b/lib/cancan/ability.rb
@@ -181,9 +181,18 @@ module CanCan
     def unauthorized_message(action, subject)
       keys = unauthorized_message_keys(action, subject)
       variables = { action: action.to_s }
-      variables[:subject] = (subject.class == Class ? subject : subject.class).to_s.underscore.humanize.downcase
+      variables[:subject] = translate_subject(subject)
       message = I18n.translate(keys.shift, variables.merge(scope: :unauthorized, default: keys + ['']))
       message.blank? ? nil : message
+    end
+
+    def translate_subject(subject)
+      klass = (subject.class == Class ? subject : subject.class)
+      if klass.respond_to?(:model_name)
+        klass.model_name.human
+      else
+        klass.to_s.underscore.humanize.downcase
+      end
     end
 
     def attributes_for(action, subject)
--- a/lib/cancan/ability.rb
+++ b/lib/cancan/ability.rb
@@ -202,6 +202,28 @@ module CanCan
       relevant_rules(action, subject).any?(&:only_raw_sql?)
     end
 
+    # Copies all rules of the given +CanCan::Ability+ and adds them to +self+.
+    #   class ReadAbility
+    #     include CanCan::Ability
+    #
+    #     def initialize
+    #       can :read, User
+    #     end
+    #   end
+    #
+    #   class WritingAbility
+    #     include CanCan::Ability
+    #
+    #     def initialize
+    #       can :edit, User
+    #     end
+    #   end
+    #
+    #   read_ability = ReadAbility.new
+    #   read_ability.can? :edit, User.new #=> false
+    #   read_ability.merge(WritingAbility.new)
+    #   read_ability.can? :edit, User.new #=> true
+    #
     def merge(ability)
       ability.rules.each do |rule|
         add_rule(rule.dup)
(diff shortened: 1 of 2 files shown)


lib/cancan/ability.rb
```diff
@@ -181,11 +181,20 @@ def authorize!(action, subject, *args)
...
       variables = { action: action.to_s }
-      variables[:subject] = (subject.class == Class ? subject : subject.class).to_s.underscore.humanize.downcase
+      variables[:subject] = translate_subject(subject)
       message = I18n.translate(keys.shift, variables.merge(scope: :unauthorized, default: keys + ['']))
       message.blank? ? nil : message
     end
 
+    def translate_subject(subject)
+      klass = (subject.class == Class ? subject : subject.class)
+      if klass.respond_to?(:model_name)
```
coorasse: in which case the class does not respond to `model_name` ? I see two cases here but only one test.

spec/cancan/ability_spec.rb
```diff
@@ -500,6 +501,19 @@ class Container < Hash
...
     end
 
+    it "uses model's name in i18n" do
+      class Account
+        include ActiveModel::Model
+      end
+
+      I18n.default_locale = :ja
+      I18n.backend.store_translations :ja,
+                                      activemodel: { models: { account: 'アカウント' } },
```
coorasse: could you please use english? You can specify it as `ja` but please use english words anyway. 
Maybe also add translations in another language `en` to see that they actually differ in your result.
```
I18n.backend.store_translations :de, activemodel: { models: { account: 'english name' } }
I18n.backend.store_translations :ja, activemodel: { models: { account: 'japanese name' } }
expect(...).to eq('japanese name...')
```

spec/cancan/ability_spec.rb
```diff
@@ -500,6 +501,19 @@ class Container < Hash
...
+                                      unauthorized: { update: { all: '%{subject}の更新権限がありません' } }
+      expect(@ability.unauthorized_message(:update, Account)).to eq('アカウントの更新権限がありません')
+      # rubocop:enable Style/FormatString
```
coorasse: this should be removed.

lib/cancan/ability.rb
nyamadori: > in which case the class does not respond to model_name ?

I think there are cases which the class does not respond to `model_name` because in following spec use Array as ability subject. So I wrote this method considering cases which doesn't respond to `model_name`.

https://github.com/CanCanCommunity/cancancan/blob/develop/spec/cancan/ability_spec.rb#L498

> I see two cases here but only one test.

Test case which does not respond to `model_name` has wrote already here:

https://github.com/CanCanCommunity/cancancan/blob/develop/spec/cancan/ability_spec.rb#L496

spec/cancan/ability_spec.rb
```diff
@@ -500,6 +501,25 @@ class Container < Hash
...
+
+      I18n.backend.store_translations :en,
+                                      activemodel: { models: { account: 'english name' } },
+                                      unauthorized: { update: { all: '%{subject}' } }
+      I18n.backend.store_translations :ja,
+                                      activemodel: { models: { account: 'japanese name' } },
+                                      unauthorized: { update: { all: '%{subject}' } }
+
+      I18n.default_locale = :en
+      expect(@ability.unauthorized_message(:update, Account)).to eq('english name')
```
coorasse: could you please use a 
```
I18n.with_locale(:fr) do
end
```

block? this avoids collateral effects and allow us to avoid the line 494

nyamadori: @coorasse Thx! I fixed:  https://github.com/CanCanCommunity/cancancan/pull/462/commits/d78398ff04ead814a96a4d97dd179d8b70b29056

nyamadori:
@coorasse I want to fix CI failures below. Is it better to suppress Rubocop offence (`Metrics/ModuleLength:
`) in Ability module or weaken the offence (e.g. Metrics/ModuleLength: 100 => 150)?

```
Offenses:
lib/cancan/ability.rb:19:3: C: Module has too many lines. [103/100]
  module Ability ...
  ^^^^^^^^^^^^^^
41 files inspected, 1 offense detected
```

https://travis-ci.org/CanCanCommunity/cancancan/jobs/306370981

coorasse:
both of those solutions are not fixes. we probably need to split the module.

coorasse:
fixed and merged. thanks

PR #464, opened 2017-11-22
update minimum ruby version in gemspec

lostapathy:
This updates the gemspec to match README as well as the build matrix travis uses.


Schwad:
Great @lostapathy 😃

PR #467, opened 2017-11-23
Test custom find_by logic

camelmasa:
For now, we didn't create unit test for `CanCan::ControllerResourceFinder#find_by_find_by_finder`.
I updated a [test](https://github.com/CanCanCommunity/cancancan/blob/develop/spec/cancan/controller_resource_spec.rb#L492-L500).
I thought that test was intended for the method.

If it is wrong, Please close it 😅


camelmasa:
@coorasse Thanks for review 😃 

> please add a new one other then replacing the existing one, otherwise we loose the previous one

I think we don't need to do that.  
https://github.com/CanCanCommunity/cancancan/blob/develop/spec/cancan/controller_resource_spec.rb#L492-L500 lines and https://github.com/CanCanCommunity/cancancan/blob/develop/spec/cancan/controller_resource_spec.rb#L511-L519 lines are tested same point. so I overwrote the test.

Issue #420, opened 2017-06-07
Define Permissions on Specific Resource Attributes


adeelejaz:
This is more of a feature request, copied from original project's 2.0 branch:

> In CanCan 2.0 it is possible to define permissions on specific resource attributes. For example, if you want to allow a user to only update the name and priority of a project, pass that as the third argument to ```can```.
> 
>   ```can :update, :projects, [:name, :priority]```
> 
> If you use this in combination with ```load_and_authorize_resource``` it will ensure that only those two attributes exist in <tt>params[:project]</tt> when updating the project. If you do this everywhere it will not be necessary to use ```attr_accessible``` in your models.
> 
> You can combine this with a hash of conditions. For example, here the user can update all attributes except the price when the product is discontinued.
> 
>  ```can :update, :products```
>  ```cannot :update, :products, :price, :discontinued => true```
> 
> You can check permissions on specific attributes to determine what to show in the form.
> 
>   ```<%= f.text_field :name if can? :update, @project, :name %>```

Considering Ryan's comment on the ticket [here](https://github.com/ryanb/cancan/issues/154#issuecomment-919114), it might be worth looking at that implementation and just export it the current project?

mohjay:
+1

pinkynrg:
+1

troyanryller:
+1

phaedryx:
This would be incredibly useful.

phaedryx:
I've been looking at the code for CanCanCan vs. CanCan 2.0 for this feature and I think it is workable. I'd be willing to work it into a pull request.

If I got this working, how soon would I be able to get it merged into a release?

phaedryx:
@coorasse I have some ideas about how to approach this, but I would love to talk it over with you. What is the best way to contact you?

coorasse:
Hello @phaedryx , I would be happy to introduce such a feature in cancancan 3.0 since is something that the community would like to see and also Ryan thought about it already seven years ago.

phaedryx:
I've got the pull request done: https://github.com/CanCanCommunity/cancancan/pull/474

Kick the tires and tell me what you think.

coorasse:
This is addressed in the branch `feature/3.0.0`. 3.0.0 is not released yet but that branch can be used in the meantime. I close the issue. Thanks everyone for your help!

pjmartorell:
[...]

PR #471, opened 2017-12-11
Changes Rule initialize method to take the more common &block style

phaedryx:
References:
https://github.com/ryanb/cancan/blob/2.0/lib/cancan/ability.rb#L109
https://github.com/ryanb/cancan/blob/2.0/lib/cancan/rule.rb#L13

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,11 +3,11 @@ cache: bundler
 sudo: false
 rvm:
   - 2.2.0
-  - 2.2.7
-  - 2.3.4
-  - 2.4.1
+  - 2.2.8
+  - 2.3.5
+  - 2.4.2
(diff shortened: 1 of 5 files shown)


phaedryx:
Note: This is a small first step towards: https://github.com/CanCanCommunity/cancancan/issues/420

coorasse:
could you please rebase on branch feature/3.0.0 and work on that one? Thanks

phaedryx:
@coorasse will do. What's the time frame for a 3.0.0 release?

coorasse:
We have nothing planned before a possible 3.0.0 release. This would be a sufficient feature to directly release as soon as it's ready. If is backward compatible (as I imagine it to be) we can release it as a minor upgrade as well

PR #474, opened 2017-12-14
Adds support for rules at the attribute/method level

phaedryx:
Allows you to add attribute-level rules, e.g.

```ruby
can :read, User, :first_name
can :read, User, :last_name
cannot :read, User, :api_key
```


phaedryx:
This isn't nearly ready to merge yet. I thought I would work through it in stages here to get feedback as I go.

phaedryx:
@coorasse something I would like you to weigh in on.

For the `can?` method Ryan added a third, `attribute` parameter here:
https://github.com/ryanb/cancan/blob/2.0/lib/cancan/ability.rb#L47

Unfortunately, in the main code base, he also added an `extra_args` parameter here:
https://github.com/ryanb/cancan/commit/69f7a659145355dbcc88635acd6e1b722432d2f3?diff=unified#diff-d4e0a70d43eb4c389e623945f417d6e0R43
in response to this issue: https://github.com/ryanb/cancan/issues/48

Which is part of the code base now. I thought it might be an undocumented or unused feature that could be replaced, but there are clearly tests for it here: https://github.com/CanCanCommunity/cancancan/blob/develop/spec/cancan/ability_spec.rb#L273

Because `extra_args` could be anything, the only way to reconcile is to change the `can?` method signature to `def can?(action, subject, attribute, *extra_args)` which would be a breaking change.

Thoughts?

coorasse:
Good point. I also didn't remember of the `extra_args` feature honestly. 😄 
I think is very useful and should be documented properly!

This brings me to a fixed point: we should preserve that feature.
Said that we have two options in front of us on how we could introduce the attributes feature: add a new parameter and break changes, add them to the `extra_args`.
in your example, if I use the extra_args I would end up with:
```
 can? :update, book, nil, [2, 3]

```
to have `[2, 3]` as extra args, which sucks.
Why don't we integrate the attributes directly in the `extra_args`?
e.g.
```
 can? :update, book, attribute: :title
```

this would not be a breaking change. what do you think?

.rubocop.yml
```diff
@@ -38,7 +38,7 @@ Lint/AmbiguousBlockAssociation:
 
 
 AllCops:
-  TargetRubyVersion: 2.0
+  TargetRubyVersion: 2.2.0
```
coorasse: Nice one. Let's extract this into a separate PR so that we can merge it already. 👍

lib/cancan/conditions_matcher.rb
```diff
@@ -1,9 +1,12 @@
 module CanCan
   module ConditionsMatcher
     # Matches the block or conditions hash
-    def matches_conditions?(action, subject, extra_args)
-      return call_block_with_all(action, subject, extra_args) if @match_all
-      return @block.call(subject, *extra_args) if @block && !subject_class?(subject)
+    def matches_conditions?(action, subject, attribute, *extra_args)
+      return call_block_with_all(action, subject, *extra_args) if @match_all
+      if @block && !subject_class?(subject)
+        return @block.call(subject, attribute, *extra_args) if attribute
```
coorasse: I'm not a fan of this syntax. I prefer the old-style `if...else` block. Nothing really important but just to let you know ;)

spec/cancan/ability_spec.rb
```diff
@@ -275,10 +281,10 @@
...
     end
 
-    expect(@ability.can?(:read, 2, 1)).to be(true)
-    expect(@ability.can?(:read, 2, 3)).to be(false)
-    expect(@ability.can?(:read, { any: [4, 5] }, 3)).to be(true)
-    expect(@ability.can?(:read, { any: [2, 3] }, 3)).to be(false)
+    expect(@ability.can?(:read, 2, nil, 1)).to be(true)
+    expect(@ability.can?(:read, 2, nil, 3)).to be(false)
+    expect(@ability.can?(:read, { any: [4, 5] }, nil, 3)).to be(true)
+    expect(@ability.can?(:read, { any: [2, 3] }, nil, 3)).to be(false)
```
coorasse: can we add a test to see how it looks with multiple `extra_args`?

coorasse:
From my point of view we could also have a breaking change and be more open to future implementations. changing the syntax to:
`can? :update, book, attribute: :title, extra_args: {...}`

CyborgMaster:
I'm very interested in having fields in our `CanCan` permissions and would love to see them as first class citizens.  

@coorasse, I agree, having to pass `nil` as a third parameter to get to the extra args is annoying.  This doesn't look very natural:
``` ruby
 can? :update, book, nil, [2, 3]
```

However, I'm not fond of having to write `attributes:` every time we check a permission.  If I fold this into our project, I'll be doing that a ton and it feels a little clunky.  It makes attributes feel tacked on and I see them as being just as important as action and subject.

What about going in between your two suggestions and making `extra_args` a named parameter and attributes an optional positional one.  That way you could have calls like this:
```ruby
 can? :update, book, extra_args: [2, 3]
```
or this:
```ruby
 can? :update, book, [:title, :author], extra_args: [2, 3]
```
because extra_args is a named parameter and the attributes are optional, both would work.  This way the three main parameters, `action`, `subject`, and `attributes` are all positional and the tack on, `extra_args` is named.  

@coorasse, what do you think?
@phaedryx, would you be willing to go this way?

.rubocop.yml
phaedryx: done!

lib/cancan/conditions_matcher.rb
phaedryx: I think rubocop's love of early returns is actually how I ended up here. ¯\_(ツ)_/¯

phaedryx:
@CyborgMaster I like the suggestion. I'll explore some code around it.

I'm interested in this because I'm working on a GraphQL API and would love to be able to easily define attribute-level permissions in a simple, straightforward way. 

@coorasse I like the symmetry of `can :action, Subject, :attribute` and `can? :action, Subject, :attribute`. Would your solution be to add `attributes:` to both?

phaedryx:
@coorasse When I run the tests locally they all pass. I'm not sure how to solve the `pg` gem error.

phaedryx:
I have removed the `nil`s from the tests where they weren't needed and wrote a test to cover two instances when they are.

That is, the `nil` is only needed when you _don't_ have attributes for a rule but you _do_ have additional arguments that look like attributes, e.g. an array of symbols `[:foo, :bar]`. I expect this is a rare edge case.

pedrocarmona:
I also need this PR. Thx @phaedryx 👍

phaedryx:
Note: because of how I handled the [initialize](https://github.com/CanCanCommunity/cancancan/pull/474/files#diff-5e5b3568dd7bbc904edf396ab78919ccR15) method of Rule, the change isn't as *breaking* as I initially thought it would be. The only incompatibility, like I mentioned above, is if you are passing a symbol or array of symbols as extra arguments (passing conditions works the same as before). The extra arguments functionality is barely remembered and undocumented, so I assume it is rarely used, let alone with an array of symbols. Adding a `nil` in this case doesn't seem like a big problem. All that said, I don't think I will add the `attributes:`.

phaedryx:
This adds a `permitted_attributes` method. I imagine it will be used like this: `params.require(:user).permit(*current_ability.permitted_attributes(:update, User))`

lib/cancan/ability/strong_parameter_support.rb
```diff
@@ -0,0 +1,25 @@
...
+      # Return an array of attributes suitable for use with strong parameters
+      def permitted_attributes(action, subject)
+        @permitted_attributes ||= {}
+        @permitted_attributes[[action, subject]] ||= allowed_attributes(action, subject)
+      end
+
+      private
+
+      def allowed_attributes(action, subject)
+        attributes = relevant_rules(action, subject).reduce([]) do |array, rule|
```
pedrocarmona: alternatively
```
attributes = relevant_rules(action, subject).flat_map do |rule|
  if rule.attributes.empty? && subject.class == Class # empty attributes is an 'all'
    subject.instance_methods.map(&:to_sym)
  else
    rule.attributes
  end
end
```

phaedryx: 👍

coorasse:
I ❤️ @CyborgMaster proposal

Appraisals
```diff
@@ -43,6 +43,6 @@ appraise 'activerecord_5.1.0' do
 
   gemfile.platforms :ruby, :mswin, :mingw do
     gem 'sqlite3'
-    gem 'pg'
+    gem 'pg', '~> 0.18'
```
coorasse: you can now rebase. this has been fixed already.

lib/cancan/ability/strong_parameter_support.rb
```diff
@@ -0,0 +1,25 @@
...
+        @permitted_attributes ||= {}
+        @permitted_attributes[[action, subject]] ||= allowed_attributes(action, subject)
+      end
+
+      private
+
+      def allowed_attributes(action, subject)
+        attributes = relevant_rules(action, subject).flat_map do |rule|
+          if rule.attributes.empty? && subject.class == Class # empty attributes is an 'all'
+            subject.instance_methods.map(&:to_sym)
```
coorasse: you are actually taking all methods and not just attributes here, or I got it wrong? in that case it smells a little bit, what do you think?

lib/cancan/rule.rb
```diff
@@ -73,12 +76,31 @@ def matches_subject?(subject)
...
+    def parse_attributes_from_extra_args(args)
+      attributes = args.shift if valid_attribute_param?(args.first)
+      extra_args = args.shift
+
+      [attributes, extra_args]
+    end
+
+    def condition_and_block_check(conditions, block, action, subject)
+      return unless conditions.is_a?(Hash) && block
+      raise BlockAndConditionsError, 'A hash of conditions is mutually exclusive with a block.'\
```
coorasse: this part could be extracted into a separate PR and immediately merged to make this PR smaller

lib/cancan/parameter_validators.rb
```diff
@@ -0,0 +1,7 @@
+module CanCan
+  module ParameterValidators
+    def valid_attribute_param?(attribute)
+      attribute.nil? || attribute.is_a?(Symbol) || (attribute.is_a?(Array) && attribute.first.is_a?(Symbol))
```
coorasse: do you want to check all of them? `(attribute.is_a?(Array) && attribute.all? {|a| a.is_a?(Symbol) })`

lib/cancan/ability/strong_parameter_support.rb
phaedryx: You're absolutely right. I should tighten it up.

lib/cancan/parameter_validators.rb
phaedryx: An artifact from copying the cancan code! You are right; I should be more thorough.

phaedryx:
I've been trying to make `extra_args` work as a named parameter and I've been having some difficulty. The primary problem is how mix conditions hashes without `{}` (which is common) with a named parameter and handle them correctly.

phaedryx:
@coorasse 
I think it is outside of the scope of this pull request, but as a follow-up I think, at a minimum, [this method](https://github.com/CanCanCommunity/cancancan/blob/develop/lib/cancan/rule.rb#L56) and [this method](https://github.com/CanCanCommunity/cancancan/blob/develop/lib/cancan/ability.rb#L189) should be renamed. Really, how strong parameters are handle and code [like this](https://github.com/CanCanCommunity/cancancan/blob/develop/lib/cancan/controller_resource_builder.rb) [and this](https://github.com/CanCanCommunity/cancancan/blob/8fc2a2dbefc542749650da6b957335e8075b34a3/lib/cancan/controller_resource_sanitizer.rb) should be reconsidered.

lib/cancan/ability/strong_parameter_support.rb
```diff
@@ -0,0 +1,25 @@
...
+      # Return an array of attributes suitable for use with strong parameters
+      def permitted_attributes(action, subject)
+        @permitted_attributes ||= {}
+        @permitted_attributes[[action, subject]] ||= allowed_attributes(action, subject)
+      end
+
+      private
+
+      def allowed_attributes(action, subject)
+        attributes = relevant_rules(action, subject).flat_map do |rule|
```
stalniy: wouldn't it be more performant to implement it using `Set` and `rule.base_behavior`? So, something like this:

```rb
attributes = relevant_rules(action, subject).inject(Set.new) do |set, rule|
  if rule.attributes.empty? && subject < ActiveRecord::Base
    attrs = subject.column_names.map(&:to_sym) - [:id]
  else 
    attrs = rule.attributes
  end

  if rule.base_behavior
     set.add(attrs)
  else
     set.subtract(attrs)
  end

  set
end
```

Eventually we get unique list of attributes and ensure that user is able to access them

stalniy: Eventually you can use `attributes.to_a` to convert them into an array

phaedryx:
@stalniy this refactor addresses your comments

coorasse:
this is proceeding well. it will take still some time but I have an eye on it 👍

phaedryx:
This branch does everything I need it to do and I've fixed all of the bugs I can think of. I consider it complete.

seanmsmith23:
Is there any reason not to use some of these features @phaedryx ? I don't think I see them documented in the wiki, but they would be useful for a project I'm working on.

coorasse:
The feature is available and can be used. Is not well documented because we are moving the wiki at the moment. Please use it! 👍

seanmsmith23:
Great, thank you!

fabianprado:
Hi, i'm using this feature (so, thanks for your work :sweat_smile:), and when I use it with one attribute works perfect:

`can? :update, @object, :my_atribute`

But when I use an array of attributes:

`can? :update, @object, [:attribute_1, :attribute_2]`

I get: "undefined method `to_sym' for #<Array:0x000055f77c97a350>."

Looking rapidly at the code, it seems like `rule.matches_attributes?` doesn't support an array.  But looking at the comments of this PR and related issues, it should work.

It's not supported yet? Is there a workaround? Or maybe I'm doing something wrong?

PR #479, opened 2018-01-29
Fixes SQL exception from rails 5.2 upgrade

lizzyaustad:
[A change in the most recent beta version of Rails 5.2](https://github.com/rails/rails/blob/213796fb4936dce1da2f0c097a054e1af5c25c2c/activerecord/lib/active_record/relation/predicate_builder.rb#L55) wraps the query attribute in a BindParam node, making the attribute's value inaccessible when it comes back to cancancan. 

If the change persists in the final version of Rails 5.2, `SubstituteBinds` could be a possible solution for avoiding the SQL error that is otherwise raised. Without this change, current active record adapter tests would fail for the 5.2.0.beta2 appraisal. I can update the gem to the non-beta version so this PR can be up to date once the final version is released if the issue still exists at that point.


jhawthorn:
> I would like to keep this PR open so that we are ready when Rails 5.2 is out of beta (I don't plan to release a version of cancancan to support Rails 5.2 until is out of beta).

5.2.0.rc1 is out if that counts :wink:

lizzyaustad:
Hi @coorasse, I went ahead and updated to 5.2.0.rc1 and added an active_record_5_adapter. I noticed @s-mage already had one implemented [here](https://github.com/CanCanCommunity/cancancan/pull/444) so I based it off of that in order to reduce conflicts. Let me know if there is anything else I should change.

tvdeyen:
5.2.0.rc2 is now out. Is there anything blocking this from getting merged?

lizzyaustad:
@coorasse updated to rc2

tvdeyen:
> release this when also final 5.2 of Rails is out

@coorasse Rails 5.2 was released today. Any plans on releasing this? Thanks

lizzyaustad:
Hi @coorasse, I've updated to the final 5.2 version of rails and added my change to the "Unreleased" section of the changelog. Let me know if there's anything else I should do. Thanks!

coorasse:
Thank you! Nice job!

PR #490, opened 2018-03-19
ActiveRecord4 Adapter: Allow can? to pass a string for Enum comparison [WIP]

enrique-guillen:
Sending the attribute name to the subject returns a string, and, on AR4 + CanCan 1.17.0, `#accessible_by` returns correct values when specifying the enum as a string.

Specs pending, and I've yet to corroborate AR4's behavior throughout its different versions (especially w.r.t. joins) -- wouldn't mind hearing feedback on the idea though.

### DIFF
--- a/lib/cancan/model_adapters/active_record_4_adapter.rb
+++ b/lib/cancan/model_adapters/active_record_4_adapter.rb
@@ -13,15 +13,17 @@ module CanCan
 
       # TODO: this should be private
       def self.matches_condition?(subject, name, value)
+        attribute_enum_string = subject.send(name)
         # Get the mapping from enum strings to values.
         enum = subject.class.send(name.to_s.pluralize)
         # Get the value of the attribute as an integer.
-        attribute = enum[subject.send(name)]
+        attribute_enum_value = enum[attribute_enum_string]
         # Check to see if the value matches the condition.
         if value.is_a?(Enumerable)
-          value.include? attribute
+          value.include?(attribute_enum_string) ||
+            value.include?(attribute_enum_value)
         else
-          attribute == value
+          [attribute_enum_value, attribute_enum_string].include?(value)
         end
       end
 


coorasse:
a test could help understanding what you want to achieve. is not clear to me, sorry

enrique-guillen:
Here's an example of the way it behaves right now:

```ruby
# Create a Shape model with a color enum attribute.
class Shape < ActiveRecord::Base
  enum color: {
    red: 0,
    green: 1,
    blue: 2
  }
end

# Create three different shapes with each color.
red = Shape.create! color: :red
green = Shape.create! color: :green
blue = Shape.create! color: :blue

# Specify that the user can read green Shapes.
# AR supports "smart" enums that can be specified with the symbol or string.
some_ability.can :read, Shape, color: :green

# The emitted SQL [probably coming from #sanitize_sql] will be:
# SELECT  `shapes`.* FROM `shapes` WHERE `shapes`.`color` = NULL  ORDER BY
# `shapes`.`id` DESC LIMIT 1
# Notice how the color was mapped to NULL, which is wrong and returns no
# objects. Would we have used `color: "green"` instead, it would have mapped it
# to 0, which is also wrong, and would have returned the incorrect color.
Shape.accessible_by(some_ability)

# Compare the green object with our conditions.
# The comparison will fail instead of succeeding, because matches_condition?
# will compare :green against Shape.defined_enums[:color][green.color], or 1.
# The same would have happened with a string.
some_ability.can? :read, green
```
[We would have to amend the procedures invoked by `accessible_by` as well.]

I understand you can reliably specify `Shape.colors["green"]` instead, especially on 4.2:

- [4.2: AR documentation says you need the ordinal value for `where` queries.](http://api.rubyonrails.org/v4.2/classes/ActiveRecord/Enum.html)
- [5.1: AR documentation says this is the "manual" way to build the SQL string.](http://api.rubyonrails.org/v5.1/classes/ActiveRecord/Enum.html)

Having said that, I suggest we support a simpler behavior here, by amending the AR4 adapters.

```ruby
some_ability.can :read, Shape, color: :green

Shape.accessible_by(some_ability) # => returns the `green` object

some_ability.can? :read, green # => true
```

coorasse:
tackled by #534 and scheduled for 2.4.0

PR #493, opened 2018-04-02
Block Rule ignored if Subject is a class

phaedryx:
Fixes bug where cannot rule should be ignored when subject is a class and no conditions and has a block


coorasse:
I am still evaluating this feature because:
1. it would mean a change in the current behavior and therefore it should be released as a major.
2. I am not sure about the expected final behavior. At the moment the rule `cannot :read, Integer { |i| i > 5 }` influnces both `Integer` as a class and single values, after this patch it would not.

I am open to receive opinions about this PR.

coorasse:
We cannot have a situation where `can? :read, Integer` returns true, meaning that you are allowed to read all integers, but that's not true for some cases. I consider that this new behaviour would be very confusing. Thanks for the PR, anyway. I appreciate all contributions.

PR #508, opened 2018-06-01
Optimised queries by using left_joins in Rails 5.2 or above

kaspernj:
Simplified version of #238 

This greatly simplifies the generated SQL queries and optimises performance, because records aren't unnecessarily preloaded when using `left_joins(...)` instead `.includes(...).references(...)`

### DIFF
--- a/lib/cancan/model_adapters/active_record_5_adapter.rb
+++ b/lib/cancan/model_adapters/active_record_5_adapter.rb
@@ -28,7 +28,15 @@ module CanCan
       # in addition to `includes()` to force the outer join.
       def build_relation(*where_conditions)
         relation = @model_class.where(*where_conditions)
-        relation = relation.includes(joins).references(joins) if joins.present?
+
+        if joins.present?
+          if Rails::VERSION::MAJOR >= 5 || Rails::VERSION::MAJOR == 5 && Rails::VERSION::MINOR >= 2
+            relation = relation.left_joins(joins)
+          else
+            relation = relation.includes(joins).references(joins)
+          end
+        end
+
         relation
       end
 


kaspernj:
The double join issue hasn't been fixed in Rails 5.2 :-/

kaspernj:
I believe I have found the issue in Rails and reported it here: https://github.com/rails/rails/issues/33048

kaspernj:
This will work, if you dont do a `load_resource through: relation_that_uses_has_many_through`

kaspernj:
I have been testing this out on my site www.peakflow.io and it works great. I had to add a couple of `.distinct` to my queries though.

In that regard Postgres complained about having to alter my SQL, but I fixed that almost seamlessly with this gem I wrote some time ago: https://github.com/kaspernj/active_record_query_fixer

coorasse:
I will schedule this for version 3.0.0. Although it keeps the previous behavior for older versions of Rails it's an important behavior change.

lib/cancan/model_adapters/active_record_5_adapter.rb
```diff
@@ -28,7 +28,15 @@ def self.matches_condition?(subject, name, value)
       # in addition to `includes()` to force the outer join.
       def build_relation(*where_conditions)
         relation = @model_class.where(*where_conditions)
-        relation = relation.includes(joins).references(joins) if joins.present?
+
+        if joins.present?
+          if Rails::VERSION::MAJOR >= 5 || Rails::VERSION::MAJOR == 5 && Rails::VERSION::MINOR >= 2
```
kaspernj: This should probably be changed to this:
```ruby
if Rails::VERSION::MAJOR > 5 || Rails::VERSION::MAJOR == 5 && Rails::VERSION::MINOR >= 2
```

kaspernj:
@coorasse What do you think about adding a configuration option to enable opt in for the new behaviour?

coorasse:
Closed in favour of #512

coorasse:
I think the new behaviour is simply the right one. The old behaviour was there just because there was no simple way to perform a `left_join`. cancancan should not eager load associations for you. the old behaviour already caused too many issues. I would just switch to it from version 3.0 👍

kaspernj:
@coorasse I would argue to remove the `.distinct`, since that can cause Postgres to crash based on the generated SQL. I had to select certain customs in the SQL in order to fix this. I don't think the average user will expect to be required to do this, unless he himself does the `.distinct`.

As I mentioned in my PR, I used this small gem to do this automatically for me: https://github.com/kaspernj/active_record_query_fixer

If you decide to keep the `.distinct`, I would urge you to consider copying the auto-fix for the query.

coorasse:
Many tests fail without distinct since it returns duplicate results. Would it be possible to create a gist to replicate the issue?

kaspernj:
@coorasse Here is the type of error, I have begun getting after trying out the 3.0 feature branch:

```
ActionView::Template::Error:
  PG::InvalidColumnReference: ERROR:  for SELECT DISTINCT, ORDER BY expressions must appear in select list
  LINE 1: ...oject_memberships"."github_user_id" = $1 ORDER BY github_use...
```

So I guess it would go something like this in order to replicate it:
```ruby
Model.accessible_by(current_ability).order(:some_column)
```

javinto:
Just wanna say that I just tried out the feature/3.0.0 branche and that the above change dramatically increases performance on my application. I hope 3.0 will get released soon. Keep up the good work!

PR #510, opened 2018-06-12
Fixed issue when doing load_resource on a has_one with the singleton option and using the current_user method in Devise

### LINKED ISSUE #58: load_and_authorize_resource through: :current_user, singleton: true
For the sake of an example I'm using a singleton profile for the current user

``` ruby
require 'spec_helper'

class Profile < Struct.new(:user); end

class ProfilesController < ApplicationController
  load_and_authorize_resource through: :current_user, singleton: true
end

describe ProfilesController do
  login_user # macro which sets :current_user for devise
  before do
    @routes.draw do
      resource :profile
    end
  end

  describe 'GET new' do
    it 'assigns a new profile to @profile' do
      get :new
      expect(assigns :profile).to be_a Profile
    end
  end
end

```

results in

``` ruby
ProfilesController GET new assigns a new profile to @profile
  Failure/Error: get :new
  NoMethodError:
    undefined method `current_user=' for #<struct Profile user={}>
```

https://github.com/CanCanCommunity/cancancan/blob/master/lib/cancan/controller_resource.rb#L90 tries to use `profile.current_user=`, but it isn't defined on the profile model.

Is there an option to override the `parent_name` to `user`

kaspernj:
Fixed #58 

I ran in to this issue while trying to do the following:
```ruby
class FreightAndTradeSettingsController < ApplicationController
  load_and_authorize_resource :freight_and_trade_setting, through: :current_user, singleton: true
```

I found it hackish to add the proposed `alias_method`. Instead I believe it would be much cleaner to reuse the already existing argument for `load_resource` called `through_association` like this:
```ruby
class FreightAndTradeSettingsController < ApplicationController
  load_and_authorize_resource :freight_and_trade_setting, through: :current_user, singleton: true, through_association: :user
```

And that is what I have done here :-)


coorasse:
due to lack of activity. I close this one. Please feel free to add a test and re-open.

PR #512, opened 2018-06-22
Kaspernj optimised queries with left joins

coorasse:
Removed eager loading of associations
Fixes CI of #508


ghiculescu:
FYI, this change causes errors in rails 5.0 and 5.1 that seem to be fixed in 5.2.

Because of this I'm proposing we only use this improvement from 5.2 onwards. https://github.com/CanCanCommunity/cancancan/pull/600#issuecomment-524672268

PR #513, opened 2018-07-01
Custom ar fix

mbasset:


mbasset:
@coorasse The ultimate problem we are having is when we have more complex logic to compare. For example when using multiple sub-queries in the matchers of the ability file. Something akin to:

```
my_employees = user.employees
can_complicated? :manage, Comments, employee_id: Permission.where(peer_id: my_employees).select(:employee_id)```

This results in a chained Active Record relation query with appropriate sub-queries which are lazy evaluated when necessary. The issue that is hard to work around with the block method currently is when using the `accessible_by(current_ability)` method. Without this change it will simply return that it is unable to evaluate these conditions. With the change it works as expected and produces the expected lazy SQL.

`Comment.accessible_by(current_ability).where(...)`

By no means is the proposed solution well designed code. However I thought it a good starting point for the discussion at minimum.

coorasse:
Please consider rebasing your fix and adding a regression test. Without further feedback, I'll close this PR. Thanks for your help!

Issue #523, opened 2018-08-15
Abilities incorrectly applied with multiple associations to the same model


kbrandwijk:
### Steps to reproduce

Given these models:
```ruby
class Account < ApplicationRecord
end

class Project < ApplicationRecord
  belongs_to :manager, class_name: 'Account'
  belongs_to :client, class_name: 'Account'
end
```
And these ability config:
```ruby
can :read, Project, manager: { name: user.name }
can :read, Project, client: { name: user.name }
```

Run this query
```ruby
Project.accessible_by(ability).first
```

### Expected behavior
I should only be able to see Projects where I am either the manager _or_ the client

### Actual behavior
The generated query only takes the first association into consideration, effectively only returning those Projects where I am the manager.
```sql
SELECT COUNT(DISTINCT "projects"."id") 
FROM "projects" 
LEFT OUTER JOIN "accounts" ON "accounts"."id" = "projects"."manager_id" 
LEFT OUTER JOIN "accounts" "clients_projects" ON "clients_projects"."id" = "projectss"."client_id" 
WHERE (("accounts"."name" = 'John') OR ("accounts"."name" = 'John'))
```
The second where clause should use the alias `clients_projects`. The correct SQL statement is:
```sql
SELECT COUNT(DISTINCT "projects"."id") 
FROM "projects" 
LEFT OUTER JOIN "accounts" ON "accounts"."id" = "projects"."manager_id" 
LEFT OUTER JOIN "accounts" "clients_projects" ON "clients_projects"."id" = "projectss"."client_id" 
WHERE (("accounts"."name" = 'John') OR ("clients_projects"."name" = 'John'))
```

### System configuration
**Rails version**:
5.2.0
**Ruby version**:
2.5.1
**CanCanCan version**
2.2.0

jrhicks:
I was able to reproduce this.

jrhicks:
Error possibly traced to this line, which always users the table name of the association and not the anticipated aliased name:

https://github.com/CanCanCommunity/cancancan/blob/8fc2a2dbefc542749650da6b957335e8075b34a3/lib/cancan/model_adapters/active_record_adapter.rb#L61

coorasse:
duplicate of #337

PR #530, opened 2018-08-31
Predict associations names

coorasse:
Rainy weekends... ⛈ 
This PR aims to solve the issues #523, #337 and close #518, #363.
Let's recap it to "the problem of the same table referenced twice".
This problem is not of CanCanCan, directly but, rather of ActiveRecord (see issues: https://github.com/rails/rails/issues/606).

**Sum up**
I cannot ask ActiveRecord to tell me the alias that will give to an association, once the SQL is generated.

**What we have tried**
Following the issues mentioned above, we tried to fix the problem first by switching to Arel and then using UNIONs.

The rewrite of CanCanCan into Arel has been excluded due to the fact that it would mean a huge rewrite and possible conflicts with ActiveRecord.

The usage of UNIONs has been experimented and had succeeded, but it caused a significative decrease of the performance.

**What does this PR do**
This PR extracts the calculation of the table aliases in a separate class which *tries* to mimic ActiveRecord behaviour when assigning an alias to a table.
This method is not bullet proof but is rather cheap in terms of execution time and solves the simplest cases. I would define it a good compromise to let the simplest situations be working.

It also adds a nice exception message when a wrong association name is used in the abilities definition.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,5 @@
+version: "2"
+checks:
+  argument-count:
+    config:
+      threshold: 5
--- a/CHANGELOG.rdoc
+++ b/CHANGELOG.rdoc
@@ -2,6 +2,8 @@ Unreleased
 
 * Compress irrelevant rules before generating a query to optimize performances (coorasse)
 * Remove ruby 2.2 from Travis and add ruby 2.5.1 (coorasse)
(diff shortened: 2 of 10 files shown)


coorasse:
@kbrandwijk could you verify if your issue is solved using this branch? Thanks

PR #533, opened 2018-10-01
Add comment in code for devise users

BKSpurgeon:


coorasse:
I don't see why devise shouldn't check for authorizations before the user logs in. Maybe the comment is wrong? You actually want to exclude devise controllers

BKSpurgeon:
I don't remember the issue exactly, but from the comment in the PR, cancancan will check for authorisation BEFORE the users have logged in - and since the users have not logged in it will throw an exception, hence, someone is using devise, they might want to hit the devise controller without authorisation.

PR #537, opened 2018-10-17
Rails api mode config

ahmad19:
This PR adds notes to the README for necessary rails API configurations.


coorasse:
The controller additions should be loaded automatically. If that's not the case, we have an issue.

coorasse:
From my tests it seems this is not necessary and the controller additions are added automatically as expected.

Issue #468, opened 2017-11-24
CanCan::Ability#merge copies only rules


npetkov:
I'm a bit confused about how the `Ability#merge` method works. Looking at the source code, the only thing it does is to transfer the rules form the argument to the receiver. If the argument defines aliased actions, they won't be copied, thus breaking the rules. I doubt this is a bug, but it nevertheless seems like inconsistent behaviour to me. If this is done by design (it may be tricky to merge alias actions), please provide some insight into the matter.

### Steps to reproduce

Please see the provided [gist](https://gist.github.com/npetkov/e7e6ff57cddacae62c19babdaeaeb7e6).

### Expected behavior

The created ability should include the aliased actions.

### Actual behavior

The created ability doesn't include anything except the rules, as this is how the `merge` method works now.

### System configuration
**Rails version**:

5.1

**Ruby version**:

2.4.1

**CanCanCan version**

2.1.1

afdev82:
Yes, I already opened this issue for that: https://github.com/CanCanCommunity/cancancan/issues/280.

Issue #280, opened 2015-12-21
Merge doesn't add aliased actions


afdev82:
I use the AlchemyCMS gem to provide the CMS functionality for a rails 4.2+ app.
I have a problem since I have to merge the permissions in the CMS with permissions defined in my app.
I already described what happens in [this issue](https://github.com/AlchemyCMS/alchemy_cms/issues/739).
It seems that the merge method of the Ability object doesn't merge the aliased actions.
Is this an intentional behaviour or is it a bug?

Haniyya:
Aliased actions are not copied over. Merge just copies rules:
```rb
    def merge(ability)
      ability.rules.each do |rule|
        add_rule(rule.dup)
      end
      self
    end
```

I agree though that this (along with a few  other things) would improve `#merge` and make it more intuitive.

coorasse:
Closed because of starvation. 
Please feel free to reopen if this issue is still valid or interested in solving with a PR.
Thank you.

coorasse:
duplicated of #468. Fixed by #538.

PR #538, opened 2018-10-22
Merge alias actions when merging abilities

Jcambass:
Closes https://github.com/CanCanCommunity/cancancan/issues/468
Closes https://github.com/CanCanCommunity/cancancan/issues/280

This changes the behaviour of `Ability#merge` to also include defined
`alias_actions`.

If a collision between the defined aliases occurs the actions on `self`
will be overwritten with those on the passed object. For an example see [this added spec](https://github.com/Jcambass/cancancan/blob/b41eb95407409a5e76f1e7c99f524a8c29fec925/spec/cancan/ability_spec.rb#L582).

The inline documentation has been updated in order to reflect those
changes.

### DIFF
--- a/lib/cancan/ability.rb
+++ b/lib/cancan/ability.rb
@@ -202,12 +202,13 @@ module CanCan
       relevant_rules(action, subject).any?(&:only_raw_sql?)
     end
 
-    # Copies all rules of the given +CanCan::Ability+ and adds them to +self+.
+    # Copies all rules and aliased actions of the given +CanCan::Ability+ and adds them to +self+.
     #   class ReadAbility
     #     include CanCan::Ability
     #
     #     def initialize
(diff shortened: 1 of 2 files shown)


lib/cancan/ability.rb
```diff
@@ -216,18 +217,43 @@ def has_raw_sql?(action, subject)
...
+    #   include CanCan::Ability
+    #
+    #   def initialize
+    #     alias_action :show, to: :see
+    #   end
+    # end
+    #
+    # read_ability = ReadAbility.new
+    # read_ability.merge(ShowAbility)
+    # read_ability.aliased_actions #=> [:see => [:show]]
```
coorasse: Honestly, I'd have expected them to be `see: [:show, :index]` since they are merged.

PR #541, opened 2018-10-22
Add assertion of `sym` be nil

yui-knk:


coorasse:
closed in favour of #557. This expectation is not executed cause the code you added never runs.

Issue #552, opened 2019-01-07
Unnecessary & slow query executed when using scope as condition


albb0920:
### Steps to reproduce
gist: https://gist.github.com/albb0920/09ff9e790d8d3d2b29dcfe98ca7f5c9e

When using a scope as condition, `conditions_empty?` in conditions_matcher.rb will check the condition against an empty hash with operator `==`, which will cause ActiveRecord to load the relation.

Code path
- "activerecord-5.2.0/lib/active_record/associations/collection_proxy.rb:46:in `load_target'"
- "activerecord-5.2.0/lib/active_record/associations/collection_proxy.rb:988:in `=='"
- "cancancan-2.3.0/lib/cancan/conditions_matcher.rb:90:in `conditions_empty?'"
- "cancancan-2.3.0/lib/cancan/rule.rb:46:in `only_block?'"
- "cancancan-2.3.0/lib/cancan/ability.rb:198:in `any?'"
- "cancancan-2.3.0/lib/cancan/ability.rb:198:in `has_block?'"
- "cancancan-2.3.0/lib/cancan/controller_resource.rb:74:in `load_collection?'"

Changing conditions_empty? as followed fixed my problem.
```
def conditions_empty?
  (@conditions.is_a?(Hash) && @conditions == {}) || @conditions.nil?
end
```

### Expected behavior
Using scope as condition should not trigger query in load_resource

### Actual behavior
Using scope as condition triggers a query in load_resource

### System configuration
**Rails version**:
5.2.0

**Ruby version**:
2.5.1

**CanCanCan version**
2.3.0

coorasse:
You should write the rule as `can :read, Book, user_id: user.id`. This way you don't need to load the association at all.

albb0920:
I believe loading records from association is unintentional,
as you can see in my description, all it want is to check whether `@conditions` is an empty hash,
which shouldn't trigger the association to load records.

The example in my gist is just to demonstrate this bug.

As for why we're use associations, 
we actually have a stored procedure in DB for authorization for a few models, that way we can share authorization logic between multiple app.
It is wrapped as a scope and used in conjunction with other scopes.

In other projects, I also find using scopes makes the ability file much easier to read.

It's just a one line fix, should I submit a PR?

albb0920:
The problem is it will load all records from DB when going through cancancan, which is very very slow.
I do pagination in my controller, this will cause it to do DB query twice.

[...]

PR #569, opened 2019-02-25
Fix accessible_by fires query for rules using association as condition.

albb0920:
see #552

In turns out this could be fixed by swapping the left hand side and right hand side of `==`.


albb0920:
Rubocop doesn't like it, do you want `(@conditions.is_a?(Hash) && @conditions == {})` instead?

spec/cancan/model_adapters/active_record_adapter_spec.rb
```diff
@@ -115,6 +115,17 @@ class User < ActiveRecord::Base
       expect(Article.accessible_by(@ability)).to eq([article])
     end
 
+    it 'does not fires query with accessible_by() for abilities defined with association' do
+      user = User.create!
+      @ability.can(:read, Article, user.articles) { true }
```
coorasse: this is not how you write rules in cancancan. you should replace it with `can :read, Article, user_id: user.id`, as discussed also #552.

albb0920:
I've updated the test case, does this look more reasonable?

coorasse:
closed in #620

PR #573, opened 2019-03-08
Fixed issue with i18n messages when using options hash (like through: :association) to load and authorize resources

bruno-campos:
When using i18n messages: https://github.com/CanCanCommunity/cancancan/wiki/Translations-(i18n)

And having an en.yml file that looks like this:
```
en:
  unauthorized:
    read:
      all: "error for %{subject}"
      post: "error for post"
```

Having the ability:
```
can :read, Blog
cannot :read, Post
```

And loading resources like:
```
load_and_authorize_resource :blog
load_and_authorize_resource through: :blog
```

When requesting the index action, it will display the error `error for hash` instead of `error for post`. 

This happens because in the method `unauthorized_message`, the subject param is a hash, and the code tries to get the Hash object class name. The hash on subject looks like this:
```
{ <Blog id:1> => Post(id:integer) }
```

This PR fixes those error messages having the unauthorize message methods look for the correct key instead of "hash" in those cases.


bruno-campos:
@coorasse since you are the active maintainer, is this something that we should fix? I came across this while playing with i18n for custom permission messages.

Issue #572, opened 2019-03-08
defining ability for HABTM 


chrisgeek:
I have the following,but it does not seem to work,is it that cancancan gem doesn't work properly with a HABTM association? the code below returns no assessments, but i have assessments related to the user in my db

``class User < ActiveRecord::Base
  has_and_belongs_to_many :assessments
end``

``class User < ActiveRecord::Base
  has_and_belongs_to_many :users
end``

```class Ability
  include CanCan::Ability
  def initialize(user)
    # a user can only manage assessments related to it 
   can :manage, Assessment, users: { id: user.id }
  end
end```

coorasse:
Hi, from #576 I can see that HABTM works just fine. If you have any issue, please fill a gist (is provided in the issue template) to reproduce it.

PR #576, opened 2019-03-16
Add explicit test for HABTM associations

coorasse:
When reading #572, I realised no explicit test for this case was present, although the system is supposed to work. This test proves the compatibility of `cancancan` with HABTM associations.


chrisgeek:
Hi i asked the opened the issue #572, Does it mean that HABTM association is not already part of this gem? @coorasse

PR #579, opened 2019-03-19
Introduce AR adapter for Rails 6 

sebjacobs:
I have ended up inheriting from the Rails 5 adapter since we should be
able to reuse most of this existing functionality, however I have had to
override `#visit_nodes` as the implementation in the Rails 5 adapter had
some conditional logic which is not relevant in Rails 6.

I initially ran into some issues getting the test suite green due to a
change in how ActiveRecord serializes sqlite booleans (it now serializes
them as integers)[1], however by extracting these hardcoded values into
some spec helpers I have been able to resolve the issue.

[1] https://www.github.com/rails/rails/commit/f59b08119b


coorasse:
Hey @sebjacobs , thanks for your PR. I released two days ago the version `3.0.0.rc1`. could you try that one? It should solve your problem. I didn't merge it yet on develop, this might have caused some confusion and work. Sorry for that 🙇 . Please give me a feedback if that version works fine for you.

sebjacobs:
> Hey @sebjacobs , thanks for your PR. I released two days ago the version `3.0.0.rc1`. could you try that one? It should solve your problem. I didn't merge it yet on develop, this might have caused some confusion and work. Sorry for that 🙇 . Please give me a feedback if that version works fine for you.

No worries at all :smile:. I must've forked `develop` before you released `3.0.0.rc1` so will close this PR and will give that release a go.

PR #590, opened 2019-05-15
Fix Rule#inspect when rule is created through a SQL array

### LINKED ISSUE #589: Rule's inspect method fails when the rule is created through a SQL array
### Steps to reproduce
Define ability with a SQL array and call #inspect.

### Example

```
CanCan::Rule.new(true, :read, Watermelon, ['visible=?', true], {}, {}).inspect
```

### Expected behavior
The method displays the rule in a string format.

### Actual behavior
There is an error:

```
 TypeError:
       no implicit conversion of nil into String
     # ./lib/cancan/rule.rb:37:in `+'
     # ./lib/cancan/rule.rb:37:in `inspect'
     ...
```

### System configuration
Rails version:
5.1.7 (not relevant)
Ruby version:
2.4.6 (not relevant)
CanCanCan version
3.0.1

frostblooded:
Fixes #589

### DIFF
--- a/lib/cancan/rule.rb
+++ b/lib/cancan/rule.rb
@@ -34,11 +34,13 @@ module CanCan
     def inspect
       repr = "#<#{self.class.name}"
       repr += "#{@base_behavior ? 'can' : 'cannot'} #{@actions.inspect}, #{@subjects.inspect}, #{@attributes.inspect}"
-      repr += if with_scope?
-                ", #{@conditions.where_values_hash}"
-              elsif [Hash, String].include?(@conditions.class)
-                ", #{@conditions.inspect}"
-              end
+
+      if with_scope?
+        repr += ", #{@conditions.where_values_hash}"
+      elsif [Hash, String].include?(@conditions.class)
+        repr += ", #{@conditions.inspect}"
+      end
+
       repr + '>'
     end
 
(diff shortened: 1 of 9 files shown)


frostblooded:
I am unsure if this spec is the best one for the case.

coorasse:
Hi @frostblooded , thanks for your PR. 👏

lib/cancan/rule.rb
```diff
@@ -34,11 +34,13 @@ def initialize(base_behavior, action, subject, *extra_args, &block)
     def inspect
       repr = "#<#{self.class.name}"
       repr += "#{@base_behavior ? 'can' : 'cannot'} #{@actions.inspect}, #{@subjects.inspect}, #{@attributes.inspect}"
-      repr += if with_scope?
-                ", #{@conditions.where_values_hash}"
-              elsif [Hash, String].include?(@conditions.class)
-                ", #{@conditions.inspect}"
-              end
+
+      if with_scope?
```
aniketstiwari: @frostblooded  I guess you have pushed this file by mistake.

frostblooded: No. This is the actual fix of the bug. The way it was previously written, when there was an SQL condition, the if/elsif wouldn't enter either of the tho blocks and it would "return" nil, which would make the code attempt to do a "repr += nil" and would throw and error.

PR #591, opened 2019-05-24
README: Typo in Ability example

olleolleolle:
A reference to a Class seems more appropriate. Speculating this is a typo. @coorasse do you know?

This PR changes the local method call to a class name.


coorasse:
Thanks!

PR #592, opened 2019-05-30
Prevent normalization of polymorphic association

### LINKED ISSUE #588: Polymorphic conditions in abilities no longer work as of 3.0.0
### Steps to reproduce

Reproduction script: https://pastebin.com/bJjq8aFU

As far as I know, CanCanCan doesn't support polymorphic associations. That is why I used a workaround (something like https://github.com/ryanb/cancan/issues/939#issuecomment-24814252 and https://stackoverflow.com/a/25966630/3659426 combined) when I want to have conditions for an ability that will check inside the polymorphic association. This used to work correctly until I updated the gem to 3.0.1 (though the issue seems to appear as of 3.0.0).

### Expected behavior
The abilities defined in the snippet to work.

### Actual behavior
Error is thrown:

```
Error:
BugTest#test_bug:
NameError: uninitialized constant Connection::Owner
    


bin/rails test cancancan_snippet.rb:71
```

### System configuration
**Rails version**:
5.1.7

**Ruby version**:
ruby 2.4.5p335 (2018-10-18 revision 65137) [x86_64-linux]

**CanCanCan version**
3.0.1

The issue seems to be occurring because of the code in `conditions_normalizer.rb` in method `calculate_result_hash` and more specifically because of what `reflection.source_reflection_name` returns.

eloyesp:
Polymorphic associations have no klass names, so those should be not
normalized. Fixes #588.

- [x] Add tests
- [x] Changelog entry

I might need help with the tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+* [#592](https://github.com/CanCanCommunity/cancancan/pull/592): Prevent normalization of through polymorphic associations.([@eloyesp][])
+
 ## 3.0.1
 
 * [#583](https://github.com/CanCanCommunity/cancancan/pull/583): Fix regression when using a method reference block. ([@coorasse][])
@@ -643,3 +645,4 @@ Please read the [guide on migrating from CanCanCan 2.x to 3.0](https://github.co
 [@andrew-aladev]: https://github.com/andrew-aladev
 [@phaedryx]: https://github.com/phaedryx
 [@kaspernj]: https://github.com/kaspernj
+[@eloyesp]: https://github.com/eloyesp
--- a/lib/cancan/model_adapters/conditions_normalizer.rb
+++ b/lib/cancan/model_adapters/conditions_normalizer.rb
@@ -31,13 +31,17 @@ module CanCan
             raise WrongAssociationName, "Association '#{key}' not defined in model '#{model_class.name}'"
           end
 
-          if reflection.options[:through].present?
+          if normalizable_association? reflection
             key = reflection.options[:through]
             value = { reflection.source_reflection_name => value }
             reflection = model_class.reflect_on_association(key)
           end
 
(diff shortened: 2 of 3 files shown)


eloyesp:
@coorasse any thought about this pull-request?

coorasse:
Hi! Thanks for your PR. 
I like your solution, since this case is currently not well managed IMHO. 
In an optimal case, it should be normalised to 
```
expect(rule.conditions).to eq(attachments: { record_id: 11, record_type: 'Article' })
```
I'd merge it by adding a TODO that explains the best scenario, not currently possible.
What are your thought about it?

eloyesp:
Let me check if I can get it working that way.

eloyesp:
@coorasse The problem with that normalization is that it depends on querying for id, it will not work on other fields.

``` ruby
    # query for id
    rule = CanCan::Rule.new(true, :read, Blob, articles: { id: 11 })
    CanCan::ModelAdapters::ConditionsNormalizer.normalize(Blob, [rule])
    expect(rule.conditions).to eq(attachments: { record_id: 11, record_type: 'Article' })

    # but does not work for other queries
    rule = CanCan::Rule.new(true, :read, Blob, articles: { status: 'draft' })
    # expect(rule.conditions).to eq(attachments: { ???? })
```

I've fixed some missing options on the sample associations and updated for rails 6.0 rc2.

Regards.

coorasse:
It could be 
```expect(rule.conditions).to eq(attachments: { record: {id: 11, type: 'Article', status: 'draft' })```?
We need to rebase, also. thanks

eloyesp:
@coorasse I've done the rebase.

> It could be
>     expect(rule.conditions).to eq(attachments: { record: {id: 11, type: 'Article', status: 'draft' })?

I'm not sure if that would work in Rails, but the main issue is that I'm not sure how to modify the actual code to do it.

PR #594, opened 2019-06-19
Add support to translate action

ayumu838:
This PR allows to translate an action's name in unauthorized message using ActiveSupport I18n.


coorasse:
Nice. Thank you.

PR #599, opened 2019-08-25
Fix a broken link

ghiculescu:


coorasse:
This README file is pretty old and old in a wrong format. It also does not contain any more information compared to what is written in the main `README.md`. We could probably simply remove it.

Issue #505, opened 2018-05-08
Negated matcher of "be_able_to" has unexpected behavior


oliver-xapix-io:
### Steps to reproduce

I forked this repo, but couldn't get the test suite running within 10 mins. I am at work, so I cannot spend too much time on this. Here's my report and monkey patch.

```
# app/models/ability.rb

can(:read, User)
``` 

```
# rspec/models/ability_spec.rb

ability = Ability.new(user)
expect(ability).to_not` be_able_to(%i[read update], User)
```

### Expected behavior

It should fail because `:read` ACTUALLY IS allowed. 

### Actual behavior

It passes.

### Monkey patch

The logic for negation over a list needs to be different.

```
# cancancan/lib/cancan/matchers.rb

Kernel.const_get(rspec_module)::Matchers.define :be_able_to do |*args|
  #...
  
  match_when_negated do |ability|                                                                                                  
    actions = args.first                                                                                                           
    if actions.is_a? Array                                                                                                         
      break false if actions.empty?                                                                                                
      actions.all? { |action| ability.cannot?(action, *args[1..-1]) }                               
    else                                                                                                                           
      ability.cannot?(*args)                                                                                                       
    end                                                                                                                            
  end

  #...
end                                                                                                                              
```

### System configuration

**Rails version**: any

**Ruby version**: any

**CanCanCan version**: any

coorasse:
You are saying "expect not to be able to read and update the User". This is logically true since you are not allowed to update, therefore you cannot do both operations. The matcher uses a AND and not a OR.

stowersjoshua:
I personally agree with @oliver-xapix-io.
Primarily because it would be the cleanest way of testing for excluded permissions.

Perhaps we can add a second matcher for this? Maybe something like `be_unable_to`? 
@coorasse - If you're open to it, I don't mind writing the PR.

stowersjoshua:
I made the mistake of writing a "fix" before checking if this had already been discussed. It would feel like a waste not to, so I [made a PR](https://github.com/CanCanCommunity/cancancan/pull/602) in case you reconsider.

PR #602, opened 2019-09-24
Implement negative expectation handler logic for `be_able_to` matcher

stowersjoshua:
The [original issue](https://github.com/CanCanCommunity/cancancan/issues/505) has already been closed. Please see my [rationale](#rationale) below.

- [x] Add tests
- [x] Changelog entry

---

Updates `be_able_to` matcher with a negative expectation handler. 

Now, if you pass an array of abilities while using `not_to`, the expectation will fail if **any** of the specified abilities turn out to be permitted.

Previously, the following test would pass, despite the user only being able to `:read` posts.

```ruby
it { is_expected.not_to be_able_to [:create, :read], Post }

...

class Ability
  include CanCan::Ability

  def initialize(_user)
    can :read, Post
  end
end
```

# Rationale
The matcher's current behavior when handling negative expectations is unlikely to be useful in many situations. Because its behavior is unexpected, it is easy to accidentally write tests that _look_ like they'll work, but then unexpectedly pass. This can be very dangerous if a developer doesn't realize how the matcher actually behaves when initially implementing the test.

The next best approach to confirm that an ability cannot be used would be to iterate over an array of abilities and test them individually. This is a bit verbose and difficult to read - especially compared to the awesome one-liner already provided.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## 3.0.2
 
+* [#505](https://github.com/CanCanCommunity/cancancan/issues/505): Implement negative expectation handler logic for `be_able_to` matcher. ([@stowersjoshua][])
 * [#590](https://github.com/CanCanCommunity/cancancan/pull/590): Fix Rule#inspect when rule is created through a SQL array. ([@frostblooded][])
 * [#592](https://github.com/CanCanCommunity/cancancan/pull/592): Prevent normalization of through polymorphic associations.([@eloyesp][])
 
@@ -650,3 +651,4 @@ Please read the [guide on migrating from CanCanCan 2.x to 3.0](https://github.co
 [@kaspernj]: https://github.com/kaspernj
 [@frostblooded]: https://github.com/frostblooded
 [@eloyesp]: https://github.com/eloyesp
+[@stowersjoshua]: https://github.com/stowersjoshua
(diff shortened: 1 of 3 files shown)


coorasse:
this stuff blows my mind all the time 😅 

```ruby
can :read, :edit, @article

is_expected.to be_able_to(:read, @article)
is_expected.to be_able_to(:edit, @article)
is_expected.not_to be_able_to(:destroy, @article)
is_expected.to be_able_to([:read, :edit], @article)
is_expected.not_to be_able_to([:read, :destroy], @article)
```

To me this is the expected behaviour. What you are proposing is
`is_expected.not_to be_able_to([:read, :destroy], @article)` should fail, because the user is able to read, and the test might be misleading.

I think we could go in circle with that and we will find always somebody that expects `is_expected.not_to be_able_to([:read, :destroy], @article)` to pass because you are not able to do both actions.

I think the problem here is how it reads and is open to interpretation. I'd rather:
* remove the possibility to pass an array
* add an alias `be_able_to_do_all` for `be_able_to`
* add a new `be_able_to_do_one_of`

This would bring us to:

```
is_expected.not_to be_able_to_do_all([:read, :destroy], @article)
```

but

```
is_expected.to be_able_to_do_one_of([:read, :destroy], @article)
```

coorasse:
I will close this issue since the current behaviour has been already discussed often in the past. I am open to a PR that creates aliases for `be_able_to` to make it simpler to use. Thanks for the contribution.

PR #605, opened 2019-11-19
Generate inner queries instead of join+distinct

### LINKED ISSUE #596: accessible_by with hash of conditions causes Not unique table/alias
### Steps to reproduce

Problem occurs with cancancan 3.x when a hash condition is defined, which
does a LEFT JOIN but the exact same table is already joined due to some eager loading or something else.

https://gist.github.com/Nghi93/c00d9dd83d3e51af5512cf4d36ed340f

### Expected behavior
In cancancan 2.x it does not do a second LEFT JOIN on the same table.

### Actual behavior
In cancancan 3.x it does a second unnecessary LEFT JOIN on the same table.

ActiveRecord::StatementInvalid: Mysql2::Error: Not unique table/alias: 'claim_files': SELECT DISTINCT `claims`.* FROM `claims` INNER JOIN `claim_files` ON `claims`.`claim_file_id` = `claim_files`.`id` LEFT OUTER JOIN `claim_files` ON `claim_files`.`id` = `claims`.`claim_file_id` WHERE `claim_files`.`policy_id` = 1 AND `claim_files`.`some_attribute` = 'a'

Maybe the LEFT OUTER JOIN could be done with an alias?
In cancancan 2.0 it seems the second join is omitted.

### System configuration
**Rails version**: 5.2.0

**Ruby version**: 2.5.5

**CanCanCan version** 3.0.1

fsateler:
When applying filters with related entities, CanCan used to generate a LEFT JOIN plus
a DISTINCT clause to avoid repeated records. This is problematic for a number of reasons:

1. Postgres is picky about setting ORDER BY with a column not on the DISTINCT clause, because it applies the ORDER after the DISTINCT so it cannot 'see' the referenced column.
2. Groups and Counts become more difficult to do because `foo.distinct.count` is not the same as `foo.count`.

The way to avoid this problem is to use inner queries that don't multiply the number of returned records. We do this by running the query we were previously using inside a WHERE IN block.

This is a different approach than #600. 

cc @ghiculescu

Fixes #596


fsateler:
I can't reproduce the CI failures locally. AFAICS, the only difference is with the postgres version (I'm using 11, CI appears to use 9.6).

fsateler:
I have reproduced this with postgres 9.6. The problem is that `selecting custom columns` and `preloading of associatons` in `accessible_by_{integration,has_many_through}` are non-deterministic since the result set contains two records and the order is not ensured. I fixed those tests by adding a condition ensuring a single result.

aleksejleonov:
@fsateler Great work!

> When applying filters with related entities, CanCan used to generate a LEFT JOIN plus
a DISTINCT clause to avoid repeated records. This is problematic for a number of reasons:

One more reason to avoid DISTINCT — it would fail when used on tables with `json` columns (PostgreSQL have no equality operator for type `json`).
`ERROR: could not identify an equality operator for type json LINE 1: SELECT DISTINCT "users".* FROM ... `

lib/cancan/model_adapters/active_record_5_adapter.rb
```diff
@@ -22,9 +22,14 @@ def self.matches_condition?(subject, name, value)
...
 
       def build_relation(*where_conditions)
-        relation = @model_class.where(*where_conditions)
-        relation = relation.left_joins(joins).distinct if joins.present?
-        relation
+        if joins.present?
+          inner = @model_class.unscoped do
+            @model_class.left_joins(joins).where(*where_conditions)
+          end
+          @model_class.where(id: inner)
```
aleksejleonov: I believe, using `@model_class.primary_key` instead of `:id` would be more safe (as primary key column could be redefined for ActiveRecord model).

```ruby
@model_class.where(@model_class.primary_key => inner)
```

coorasse:
I really like this PR! 👍 I'd like to run a couple of tests locally before merging it. Thanks for the nice work!

derosm2:
Any update on this? Excited to see this get merged in.

coorasse:
I am running the last tests and I included also #608 . Please see #619 . Will be released soon. Very nice work community! ❤️ you.

coorasse:
closed in #619

spec/cancan/model_adapters/accessible_by_integration_spec.rb
```diff
@@ -76,7 +76,7 @@ class Editor < ActiveRecord::Base
 
   describe 'preloading of associatons' do
     it 'preloads associations correctly' do
-      posts = Post.accessible_by(ability).includes(likes: :user)
+      posts = Post.accessible_by(ability).where(published: true).includes(likes: :user)
```
ghiculescu, line 79: i don't think these changes had any affect on tests?

fsateler: It did. See the explanation https://github.com/CanCanCommunity/cancancan/pull/605#issuecomment-556009976

The query changed just enough to get postgres to return a different record (because the query contained 2 but we applied limit 1).

ghiculescu: Ahhh yep right you are. I was only testing locally in sqlite when I wrote that comment. It seemed to not make a difference there.

PR #608, opened 2019-12-06
Spec for json column regression

aleksejleonov:
Starting https://github.com/CanCanCommunity/cancancan/commit/4d1f0d49bbe013a013efd95bbc50cfbd4ff2f697 `.distinct` used in `build_relation` (if joins present).

DISTINCT would fail when used on tables with json columns (PostgreSQL have no equality operator for type json).
Example: `ERROR: could not identify an equality operator for type json LINE 1: SELECT DISTINCT "users".* FROM ...`

Looks like this is regression, so added spec to cover this case.
This PR intentionally fails (TravisCI) on develop branch for postgresql adapter (due to regression).


aleksejleonov:
#605 could be a solution.

coorasse:
closed in #619

PR #618, opened 2020-02-26
Add support for using an empty array as a condition

ghiculescu:
I want to define an ability check that only applies if a `has_many` relation returns an empty array. The use case is I want to only allow deleting a record if it has no associated records for a given association.

This PR implements that functionality. Note that it only works for `can?` checks, not `accessible_by`; there's a comment explaining why. I'd like to add a warning to https://github.com/CanCanCommunity/cancancan/wiki/defining-abilities that explains this too if this is approved.

### DIFF
--- a/spec/cancan/model_adapters/active_record_adapter_spec.rb
+++ b/spec/cancan/model_adapters/active_record_adapter_spec.rb
@@ -411,6 +411,23 @@ WHERE "articles"."published" = #{false_v} AND "articles"."secret" = #{true_v}))
     expect(Article.accessible_by(ability)).to eq([article])
   end
 
+  it 'allows an empty array to be used as a condition' do
+    a1 = Article.create!
+    a2 = Article.create!
+    a2.comments = [Comment.create!]
+
+    @ability.can :read, Article, comment_ids: []
+
+    expect(@ability.can?(:read, a1)).to eq(true)
+    expect(@ability.can?(:read, a2)).to eq(false)
+
+    # ideally this would return [a1]. it doesn't because in ActiveRecord5Adapter#build_relation
+    # we call `@model_class.where(*where_conditions)`, and `@model_class.where(:comment_ids=>[]).to_sql`
+    # returns "SELECT \"articles\".* FROM \"articles\" WHERE 1=0". a proper solution would probably require
+    # left outer joins support in cancancan.
+    expect(Article.accessible_by(@ability)).to eq([])
+  end
+
   context 'with namespaced models' do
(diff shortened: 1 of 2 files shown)


ghiculescu:
@coorasse so I know if it's worth going further with this approach. Is this PR likely to get merged?

coorasse:
Hi @ghiculescu , I like your idea but having a different behaviour between `can? :read, @article` and `Article.accessible_by` scares me. Especially since we are using a hash syntax here.
I'd rather add a counter cache column and us it, if we cannot solve it also for `accessible_by`

spec/cancan/model_adapters/active_record_adapter_spec.rb
```diff
@@ -443,6 +443,71 @@ class User < ActiveRecord::Base
     expect(Article.accessible_by(ability)).to eq([article])
   end
 
+  it 'allows an empty array to be used as a condition for a has_many, but this is never a passing condition' do
```
ghiculescu, line 446: this test passes on `master`, without any of the changes in this PR

ghiculescu:
@coorasse I've tidied up the PR body and removed the one inconsistent use case. This now makes cancan behaviour match ActiveRecord behavior. Can you please take another look?

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * [#571](https://github.com/CanCanCommunity/cancancan/pull/571): Allows to check ability even the object implements `#to_a`. ([@mtsmfm][])
 * [#612](https://github.com/CanCanCommunity/cancancan/pull/612): Suppress keyword arguments warning for Ruby 2.7.0. ([@koic][])
 * [#569](https://github.com/CanCanCommunity/cancancan/pull/569): Fix accessible_by fires query for rules using association as condition. ([@albb0920][])
+* [#618](https://github.com/CanCanCommunity/cancancan/pull/618): Allow using an empty array as a condition. ([@ghiculescu][])
```
ghiculescu, line 8:
```suggestion
* [#618](https://github.com/CanCanCommunity/cancancan/pull/618): Add support for using an nil relation as a condition. ([@ghiculescu][])
```

coorasse (commented):
could you please add a test for expectations on the generated SQL query? You can find similar ones in active_record_adapter_spec.rb

ghiculescu:
Thanks @coorasse that's done :)

ghiculescu:
https://github.com/CanCanCommunity/cancancan/pull/653

PR #619, opened 2020-03-06
Use inner queries instead of left join

coorasse:
This PR combines 
* https://github.com/CanCanCommunity/cancancan/pull/608
* https://github.com/CanCanCommunity/cancancan/pull/605

and updates the CHANGELOG accordingly.


claudiosv:
This broke a bunch of stuff for us in the form of
```ruby
 @contacts = @contacts
        .where(messages: { sender_id: params[:sender_id] })
        .distinct_contact_id #with or without our distinct scope
```
where now a `.join(:message)` is needed.

Would've been nice to have a warning in the changelog as I didn't realize the implication of cancancan's internal change 🙂

coorasse:
this could actually be considered a breaking change. 🤔

ghiculescu:
should we update https://github.com/CanCanCommunity/cancancan/wiki/Migrating-from-CanCanCan-2.x-to-3.0 if we no longer use `distinct`?

ghiculescu:
Also, this PR has the potential to cause some interesting performance problems. We just struck one.  Here's the essence of the issue.

Say a user belongs to an organisation, and you have an ability `can :read, User, organisation_id: whatever`. In a controller you're doing `User.accessible_by().joins(:organisation)` in your code. Before this change, the generated SQL would look like:

```ruby
User.joins(:organisation).left_joins(:organisation).to_sql
"SELECT \"users\".* FROM \"users\" INNER JOIN \"organisations\" ON \"organisations\".\"id\" = \"users\".\"organisation_id\""
```

After this it looks like:

```ruby
User.joins(:organisation).where(id: User.left_joins(:organisation)).to_sql
"SELECT \"users\".* FROM \"users\" INNER JOIN \"organisations\" ON \"organisations\".\"id\" = \"users\".\"organisation_id\" WHERE \"users\".\"id\" IN (SELECT \"users\".\"id\" FROM \"users\" LEFT OUTER JOIN \"organisations\" ON \"organisations\".\"id\" = \"users\".\"organisation_id\")"
```

This could get a totally different query plan. In our case, the switch from old to new resulted in a primary key index not being used for the subquery, and a full table scan happening instead. Which ground everything to a halt.

I don't have any good ideas yet on how to fix this inside cancan (at our level we can tweak our indexes or our queries) but flagging it here for others.

coorasse:
Thank you for reporting this! I personally start thinking more and more that there should be an optional attribute to decide the strategy of the query for `accessible_by` (left joins, union, or inner queries). leave one option as default and allow to switch to another in case of performance issues. In bigger projects we also face such issues.

ghiculescu:
Yes, I agree. I think this is definitely a safer default, but it can cause very big and hard-to-test-for regressions in established codebases. I like the idea of strategies. I can make a PR.

coorasse:
I'd love to review and merge it if you have time for that. in the previous PRs you will find how the previous versions looked like ❤️

ghiculescu:
> I'd love to review and merge it if you have time for that. in the previous PRs you will find how the previous versions looked like ❤️

here's a draft https://github.com/CanCanCommunity/cancancan/pull/655

23tux:
We also have a huge problem with the inner queries, consider this rule:

```
can :index, Spree::Product, vendor: { active: true }
```

Doing a `Product.accessible_by(ability).to_a` results in this SQL query:

```
SELECT `spree_products`.* FROM `spree_products` WHERE `spree_products`.`id` IN (SELECT `spree_products`.`id` FROM `spree_products` LEFT OUTER JOIN `spree_vendors` ON `spree_vendors`.`id` = `spree_products`.`vendor_id` WHERE `spree_vendors`.`active` = TRUE)
```

The problem is: `spree_products` has millions of entries, so it results in HUGE DB queries.

I think the only quick option is downgrading CanCanCan :(

ghiculescu:
@23tux do you want to try this branch and see if it works for you? https://github.com/CanCanCommunity/cancancan/pull/655

Set `CanCan.accessible_by_strategy = :left_join` in an initializer.

23tux:
@ghiculescu thanks for the PR, this seems to work. This SQL is generated:

```
SELECT DISTINCT `spree_products`.* FROM `spree_products` LEFT OUTER JOIN `spree_vendors` ON `spree_vendors`.`id` = `spree_products`.`vendor_id` WHERE `spree_vendors`.`active` = TRUE
```

Can you explain:

1. Why is it a `LEFT OUTER JOIN` and not an `INNER JOIN`?
2. Isn't the use of `DISTINCT` something the application that users CanCanCan should decide? Whats the idea behind guaranteed unique results? I know this is part of CanCanCan itself and not your PR, I just like to understand the reasoning behind it

ghiculescu:
@23tux that's the query cancan generated prior to version 3.1.0; it was changed in this PR ([changelog](https://github.com/CanCanCommunity/cancancan/compare/3.0.2...3.1.0#diff-17131cd8d61b3445e02144eac7a7ed2c3215bf4b16f41875d665cbb4693e624aL534)). I'm not sure if you were using cancan before then, but my PR just tries to bring back an option for the old behavior.

That said, I'll have a go at answering your questions beyond just "that's how it always was".

1. It's possible to define `can` statements where there are no associated records. See https://github.com/CanCanCommunity/cancancan/pull/653 - `accessible_by` already works with that `can` syntax; it's just `can?` that doesn't (and that's what that PR is fixing). There's probably other/better reasons too, but this one is just one I have fresh in mind. @coorasse might know more.
2. I think it would be surprising to have `accessible_by` return multiple instances of the same record. It was added in https://github.com/CanCanCommunity/cancancan/pull/238 when eager loading was removed (I guess rails `includes` handles distincting for you), but was removed shortly afterwards and replaced with the subquery strategy.

I'd be down to add a `inner_join` option to #655 if you're interested in building it.

Issue #607, opened 2019-12-02
Please Change Privileges for editing wiki pages


BKSpurgeon:
### What is the problem?
Looks like there is no ability for non-callorators to edit the wiki pages. See here for example: https://github.com/CanCanCommunity/cancancan/wiki/Nested-Resources - I am unable to amend the wiki documentation.

### What's wrong with that?

It places additional burdens on the maintainers of the gem to update the documentation. A lot of people can assist in fixing small issues here and there if they can edit the wiki.

### How to fix:

Please change access permissions for wiki pages. See here: https://help.github.com/en/github/building-a-strong-community/changing-access-permissions-for-wikis

coorasse:
This has happened because the Wiki was completely out of control and became a huge mess. I'd be happy to move all the documentation within the repository in a `docs` folder, so that people can help maintaining the documentation by opening PR. If you feel like, you could help us by moving one page per time. This would help a lot ❤️

BKSpurgeon:
I've created a separate PR for this with all the wiki documents within the docs folder. Please note the PR noted above.

PR #621, opened 2020-03-06
Add: wiki documentation into a separate docs folder

BKSpurgeon:
Solves #607

> This has happened because the Wiki was completely out of control and became a huge mess. I'd be happy to move all the documentation within the repository in a docs folder, so that people can help maintaining the documentation by opening PR. If you feel like, you could help us by moving one page per time. This would help a lot heart

I have moved all the wiki docs into a separate folder called docs with this PR.


coorasse:
Thanks for the work. I'll start soon merging and moving.

Issue #617, opened 2020-02-20
Checking Abilities - Checking with Class - THIS IS REALLY REALLY WRONG!


Pepan:
### Steps to reproduce

Read wiki https://github.com/CanCanCommunity/cancancan/wiki/Checking-Abilities#checking-with-class or use it in your code ...

### Expected behavior
I personally prefer just pass class to block, let's developer deal with it. It is much better than just treat everything as **true** ... it leads to unsolvable problems.

#### For example
There is conditioning block that user manager of "something" is able to access items of this "something" (_you are not able to specificities it without knowing "something"!!! so you cannot have condition before **can**_)
``` Ruby
    can %i[read edit update destroy], Company do |company|
      @user.has_role? :manager, company&.account
    end
```  

this leads to whatever user asks for ```can? :index, Company``` will get response **true** ... what the heck? For example *guest*!!! 

Does developer said that? No! So why should *guest* have access to code after this last ```can?``` ????

### Actual behavior
```can? :index, Company``` responds true for any user who is asking even if it is not specified in Abilities.
there is nothing like 
``` Ruby
    can %i[read edit update destroy], Company if user.present?
```  
just
``` Ruby
    can %i[read edit update destroy], Company do |company|
      @user.has_role? :manager, company&.account
    end
```  
which have condition inside which should prevent anyone to access records except one who passed the block with true

### System configuration
**Rails version**: Rails 6.0.2.1

**Ruby version** :ruby 2.6.5-p114

**CanCanCan version**: 3.0.2

Pepan:
related issue https://github.com/CanCanCommunity/cancancan/issues/401 - with weird answer ... I will do monkey patch - but why every one should want reliable system must monkeypatch???

problem is this method where `@base_behavior` is hardcoded as **true**
/ruby-x.x.x/gems/cancancan-3.0.2/lib/cancan/conditions_matcher.rb:21
```
    def matches_block_conditions(subject, *extra_args)
      return @base_behavior if subject_class?(subject)

      @block.call(subject, *extra_args.compact)
    end
```
not clever :/

coorasse:
Calm down and open a pull request like everyone else or use the magic fork button and make your `reallyreallyrightcancancan` gem. 
There is no space for wining here, only people putting a lot of time and effort into maintaining a library used by thousands. 
The fact that you cannot understand it doesn't make it REALLY REALLY WRONG.

Pepan:
[...]

PR #624, opened 2020-03-18
Disable to pass rule with class subject with block or condition

Pepan:
fixing weird behaviour, when using block or condition in rule with class subject enables this rule for everyone, who is asking with class subject

see https://github.com/CanCanCommunity/cancancan/issues/617 for details (Actual behaviour section)

### Fixed behaviour 

```Ruby
    can(%i[read edit update destroy], Company) { |company| @user.has_role?(:manager, company.account) }
    can :index, Company if @user.roles.find_by(name: :manager, resource_type: 'Account').present?
```
- If there is only first rule, ```can? :index, Company``` will respond with **false**
- If there is also second rule, ```can? :index, Company``` will respond with **true** (if user is manager of some account)

### DIFF
--- a/lib/cancan/conditions_matcher.rb
+++ b/lib/cancan/conditions_matcher.rb
@@ -19,7 +19,8 @@ module CanCan
     end
 
     def matches_block_conditions(subject, *extra_args)
-      return @base_behavior if subject_class?(subject)
+      # we cannot process block for class subject (block is working with instance), so if we don't know what is result of block we are returning false
+      return false if subject_class?(subject)
 
       @block.call(subject, *extra_args.compact)
     end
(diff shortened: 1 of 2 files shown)


ghiculescu:
I'm not sure I agree with this change. This is clearly documented: https://github.com/CanCanCommunity/cancancan/wiki/Checking-Abilities#checking-with-class

image

And it feels like we are removing valid behavior by changing it.

ghiculescu:
> ### Fixed behaviour
> ```ruby
>     can(%i[read edit update destroy], Company) { |company| @user.has_role?(:manager, company.account) }
>     can :index, Company if @user.roles.find_by(name: :manager, resource_type: 'Account').present?
> ```
> 
> * If there is only first rule, `can? :index, Company` will respond with **false**
> * If there is also second rule, `can? :index, Company` will respond with **true** (if user is manager of some account)

The correct way to write the abilities here would be:

```ruby
if @user.roles.find_by(name: :manager, resource_type: 'Account').present?
  can(%i[read edit update destroy], Company) { |company| @user.has_role?(:manager, company.account) }
end
```

(note `index` is an [alias](https://github.com/CanCanCommunity/cancancan/blob/6d0e62bf0c475964a0533608e7e7221be45de28c/lib/cancan/ability/actions.rb#L56) of `read`)

So if the user is a manager, they can read Companies.
If the user is a manager of a specific Company, they can read that company.

coorasse:
I agree with @ghiculescu . Thanks for your effort 🙏

PR #629, opened 2020-04-16
Updated README for appraisal command.  

jpgeek:
Updated to a supported version of rails (for serious security fixes).
Added 'bundle exec' to command.
Fixed version syntax.  Appraisals file has no `activerecord_5.0` line. It looks like it specifies down to the minor version:
`appraise 'activerecord_5.2.2' do`


jpgeek:
Found some other changes that should probably be here.  Will update and resubmit.

PR #639, opened 2020-06-09
Add #inspect on AccessDenied error

eloyesp:
Access denied errors use the message attribute to help the end user (not the developer) making debugging those errors trickier, as most information is not there.

This PR adds a custom `inspect` method intended for developers, with lot of debugging information.


coorasse:
Nice! Thanks!

PR #641, opened 2020-06-29
Update before_filter to before_action in docs

alexanderweiss:
Seems like this has been missed ☺️ (it's been updated in other places).


kellydaniels:
I was going to submit the same change, but found this PR. There is also a reference to before_filter in [this file](https://github.com/CanCanCommunity/cancancan/blob/61530988c20a3d93e2ef7f598879f89582d2df34/docs/Ensure-Authorization.md) that I imagine needs to be updated as well.

coorasse (approved):

Issue #645, opened 2020-08-24
be_able_to negative matcher passes with accessable actions


mraidel:
### Steps to reproduce

```
it { expect(ability).not_to be_able_to(%i[update destroy], Article) }
```
If "update" is forbidden but destroy is not, the line above will still pass. 

### Expected behavior

I would expect the negative form of be_able_to to fail if there is an allowed action given to the matcher.

### Actual behavior
Currently it is enough to have not forbidden action per negative invocation of be_able_to. There is even a spec covering this behavior. We still think it is wrong because when I want to test if a certain set of actions is forbidden, I consider it a false positive if the spec passes while actually there are allowed actions in the list given to the matcher.

**CanCanCan version**

3.1.0

mraidel:
#646

PR #646, opened 2020-08-24
negated matcher for be_able_to only passes when no ability is eligible

mraidel:


mraidel:
Issue with description: #645

23tux:
What's the state of this PR? Can it be merged?

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## Unreleased
+
+* [#645](https://github.com/CanCanCommunity/cancancan/issues/645): Generate inner queries instead of join+distinct. ([@mraidel][], [@gabsi20][])
```
ghiculescu, line 3: this needs to be changed

ghiculescu:
```suggestion
* [#645](https://github.com/CanCanCommunity/cancancan/issues/645): Fix bug in `be_able_to` matcher. ([@mraidel][], [@gabsi20][])
```

maybe this?

mraidel:
@ghiculescu thanks for the comment, I have updated the PR with a squashed commit including the correct commit message in the changelog, also resolved the merge conflict.

coorasse:
related to #505

coorasse:
I think we discussed this issue too often. But something changed recently: I found out this behaviour in RSpec, you can define:

```
    expect([1, 3, 5, 7]).to all(be_odd)
    expect([1, 3, 5, 7]).not_to all(be_even)
```

this is very similar to what we are discussing here. Guess what does RSpec do? It raises

```
NotImplementedError: `expect().not_to all( matcher )` is not supported.
```

I think we should do the same

coorasse:
Sorry @mraidel , I was in a rush (and maybe a bit rude 😅 ) in my comment above. 

I really appreciate the effort you did put in this PR 🙏 . I'd love to get your feedback regarding my proposal.

mraidel:
Hi @coorasse, I understand the RSpec behavior for "all" because it mirrors the use of the language, "not all" just doesn't mean "none". I don't see the same issue in ".not_to be_able_to" because in this case for me the new behavior also mirrors the language. If I say "I don't want to be able to do these things" I would expect that I can do none of these, right? An alternative solution would be to disallow the negated matcher and have a new matcher "be_unable_to". Maybe there is still some possible ambiguity because one could interpret it either as "not be able to do a and b" or "not to be able to do a or b". In this case the only solution I would see is to disallow using the negated matcher with more than one argument. Just disallowing the negated matcher would miss any way to test for someone unable to access something which I think is a quite important thing to test.

coorasse:
I would disallow it only with multiple actions. You would still be able to run it with a single action.

Issue #532, opened 2018-09-14
Using STI with cancan


lcjury:
### Steps to reproduce

Hi, We're using Rails STI, so, for example, let's say we have the following models: `Survey`, `Climate` and `Evaluation`. `Climate` and `Evaluation` inherit from `Survey`. (Survey has a type column to make STI work)

When you define an ability for `Survey`, those abilities are inherited by `Climate` and `Evaluation`, but, when I define an ability for a child class (Climate) and I try to use the `accessible_by` method from the parent class, the child class abilities are ignored.

Let's say this is in my ability:

```
can :destroy, Evaluation
```

If I try to do the following:

```
Survey.accessible_by(ability, :destroy)
```

it returns nothing. I expect it to return all the Surveys of type Evaluation (or all the Evaluations). The following works:
(let's say the first Survey is a Climate and the second a Evaluation)

```
ability.can? Survey.first, :destroy #false
ability.can? Survey.second, :destroy #true
```

Gist: https://gist.github.com/lcjury/e13f706f33010d3cb80c58e87fbe5706

### Expected behavior
`Survey.accessible_by(ability, :destroy)` should return all the Surveys where `ability.can? survey, :destroy` whould evaluate to true

### Actual behavior
`ability.can? Survey.second, :destroy #true` is true, but it's not returned by `Survey.accessible_by(ability, :destroy)`

### System configuration
**Rails version**:
5.0
**Ruby version**:
2.5
**CanCanCan version**
1.17

Note: I have troubles running the gist, I get the following message: "CanCan::NotImplemented: This model adapter does not support fetching records from the database.". So, I wasn't able to test that the last lines trully works as I have told here, but I tested everything inside a running rails project.

Liberatys:
cc @coorasse As far as I can tell this issue was address by #649

coorasse:
Yes, this can be closed since is released in 3.2.0

PR #648, opened 2020-08-25
Attempt on fixing sti support

Liberatys:
Alter sti_normalizer to only create a new rule when the base class differs from the current class.

Issue: #532


Liberatys:
There seem to be multiple issues with my try on the issue..

In some tests: An error occurred while installing nokogiri (1.10.10), and Bundler fails
And in others JRuby is not displaying the same behaviour as Ruby 2.5.1 <=

PR #649, opened 2020-08-29
Add support for single table inheritance

Liberatys:
Implement a detection for sti and filter issue are caused by inputs not being validated


lib/cancan/ability.rb
```diff
@@ -302,7 +302,11 @@ def extract_subjects(subject)
 
     def alternative_subjects(subject)
       subject = subject.class unless subject.is_a?(Module)
-      [:all, *subject.ancestors, subject.class.to_s]
+      if subject.methods.include? :subclasses
```
coorasse, line 305: when does it happen that it does not have the subclasses method? can you use `if subject.respond_to?(:subclasses)`?

lib/cancan/model_adapters/sti_normalizer.rb
```diff
@@ -0,0 +1,32 @@
+# this class is responsible of normalizing the hash of conditions
+# by exploding has_many through associations
+# when a condition is defined with an has_many thorugh association this is exploded in all its parts
```
coorasse, line 3:
```suggestion
# when a condition is defined with an has_many through association this is exploded in all its parts
```

lib/cancan/model_adapters/sti_normalizer.rb
```diff
@@ -0,0 +1,32 @@
+# this class is responsible of normalizing the hash of conditions
```
coorasse, line 1: I think this part of comments should be removed. it has been copy-pasted from the rules normalizer.

lib/cancan/model_adapters/sti_normalizer.rb
```diff
@@ -0,0 +1,32 @@
...
+              true
+            end
+            subjects.length == rule.subjects.length
+          end
+          rules_cache.each { |rule| rules.push(rule) }
+        end
+
+        private
+
+        def build_new_rule(rule, subject)
```
coorasse, line 25: what about `build_rule_for_superclass`? I think it would make it more clear what the method does. a comment wouldn't hurt as well

lib/cancan/rule.rb
```diff
@@ -125,3 +137,28 @@ def wrap(object)
     end
   end
 end
+
+class SubjectClassMatcher
```
coorasse, line 141: Can you explain me why is this needed? If we are already building all the new rules. Please move it into a file itself with some comments about what is it doing and why is necessary

spec/cancan/model_adapters/active_record_adapter_spec.rb
```diff
@@ -644,4 +644,66 @@ class JsonTransaction < ActiveRecord::Base
...
+      end
+
+      class Motorbike < Vehicle
+      end
+    end
+
+    it 'recognises rules applied to the base class' do
+      u1 = User.create!(name: 'pippo')
+
+      brand = Brand.create!
```
coorasse, line 683: why do you need the Brand model? 🤔

spec/cancan/model_adapters/active_record_adapter_spec.rb
```diff
@@ -644,4 +644,66 @@ class JsonTransaction < ActiveRecord::Base
...
+      car = Car.create!(brand: brand)
+      motorbike = Motorbike.create!(brand: brand)
+
+      ability = Ability.new(u1)
+      ability.can :read, Vehicle
+      # puts Vehicle.accessible_by(ability)
```
coorasse, line 689: please remove commented code

coorasse (commented):
I think you did a great job here and I am really looking forward to push this feature forward. Before doing that I have some questions where probably I need some explanation(s). Also, I'd like to see a test for 3 levels of abstraction (it should work from what I read. Maybe add a `Suzuki` class that extends the Motorbike. Also some tests on `be_able_to` would be good.

lib/cancan/ability.rb
Liberatys: Some of the subjects are Ranges which do not have subclasses

lib/cancan/rule.rb
Liberatys: It is used to match subclasses and to upmatch from children to their ancestors to look for sti relationships

coorasse (approved):

Issue #536, opened 2018-10-17
hash_condition_match returns nil when subject is ActiveRecord Collection Proxy


BookOfGreg:
When the subject is an `ActiveRecord_Associations_CollectionProxy(ActiveRecord::Relation)`, the inner_foo value is true, but the outer_foo value is nil as seen below;

This is because it calls this method instead of array's `any?`:
https://api.rubyonrails.org/classes/ActiveRecord/Associations/CollectionProxy.html#method-i-any-3F

<img width="813" alt="screen shot 2018-10-17 at 17 10 56" src="https://user-images.githubusercontent.com/637868/47100617-d98d1d80-d22f-11e8-8c3f-9512f2cf16f2.png">

### System configuration
**Rails version**:
5.2.0

**Ruby version**:
2.5.1

**CanCanCan version**
2.3.0

BookOfGreg:
I'm still struggling to isolate exactly why `any?` is returning nil, but while debugging it, got a casual segfault from the gem which makes no sense to me so far. Putting here incase I manage to debug it in the next couple days.

```
/Users/ubuntu/source/cancancan/lib/cancan/conditions_matcher.rb:75: [BUG] Segmentation fault at 0x0000000000000004
ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-darwin16]

-- Crash Report log information --------------------------------------------
   See Crash Report log file under the one of following:
     * ~/Library/Logs/DiagnosticReports
     * /Library/Logs/DiagnosticReports
   for more details.
Don't forget to include the above Crash Report log file in bug reports.

-- Control frame information -----------------------------------------------
c:0133 p:0014 s:0920 E:0009f8 BLOCK  /Users/ubuntu/source/cancancan/lib/cancan/conditions_matcher.rb:75 [FINISH]
c:0132 p:---- s:0916 e:000915 IFUNC
c:0131 p:---- s:0913 e:000912 CFUNC  :each
c:0130 p:0018 s:0909 E:0018e0 METHOD /Users/ubuntu/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/activerecord-5.2.0/lib/active_record/relation/delegation.rb:41 [FINISH]
c:0129 p:---- s:0901 E:000070 CFUNC  :any?
[...]

PR #650, opened 2020-08-31
Feature/536 investigate association rule issue

Liberatys:
Fix the issue outlined in #536 and validate behaviour in test

### DIFF
--- a/lib/cancan/conditions_matcher.rb
+++ b/lib/cancan/conditions_matcher.rb
@@ -78,7 +78,8 @@ module CanCan
 
     def hash_condition_match?(attribute, value)
       if attribute.is_a?(Array) || (defined?(ActiveRecord) && attribute.is_a?(ActiveRecord::Relation))
-        attribute.any? { |element| matches_conditions_hash?(element, value) }
+        match_results = attribute.map { |element| matches_conditions_hash?(element, value) }
+        match_results.any?
       else
         attribute && matches_conditions_hash?(attribute, value)
       end
--- a/lib/cancan/conditions_matcher.rb
+++ b/lib/cancan/conditions_matcher.rb
@@ -78,8 +78,7 @@ module CanCan
 
     def hash_condition_match?(attribute, value)
       if attribute.is_a?(Array) || (defined?(ActiveRecord) && attribute.is_a?(ActiveRecord::Relation))
-        match_results = attribute.map { |element| matches_conditions_hash?(element, value) }
-        match_results.any?
+        attribute.to_a.any? { |element| matches_conditions_hash?(element, value) }
       else
         attribute && matches_conditions_hash?(attribute, value)
       end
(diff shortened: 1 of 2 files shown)


coorasse (commented):
does your PR allows us to use models directly instead of ids? could you please add some more complex tests and maybe verify the consistency of this behaviour also with `accessible_by` method?

Liberatys:
Sure ^^ I will see to it today or tomorrow :D

Liberatys:
I added some tests to verify that this also works with `accessible_by`.
Also added an example with multiple users and models that validate the access is made based on the role of a user

lib/cancan/conditions_matcher.rb
```diff
@@ -78,7 +78,8 @@ def condition_match?(attribute, value)
 
     def hash_condition_match?(attribute, value)
       if attribute.is_a?(Array) || (defined?(ActiveRecord) && attribute.is_a?(ActiveRecord::Relation))
-        attribute.any? { |element| matches_conditions_hash?(element, value) }
+        match_results = attribute.map { |element| matches_conditions_hash?(element, value) }
+        match_results.any?
```
ghiculescu, line 82: rather than the multi-line solution, how do you feel about `attribute.to_a.any? { |element| matches_conditions_hash?(element, value) }`

if `attribute` is already an array, `to_a` will do nothing. if it's not, it'll force the relation to be loaded.

i just think that code makes a bit more sense than what you have here - if i saw this, i'd want to refactor it back to the old version.

Liberatys: That absolutely makes a lot of sense ^^ Your change looks a lot cleaner and thus I'm happy to revert to just adding to_a to force the relation loading

coorasse (approved):

lib/cancan/conditions_matcher.rb
```diff
@@ -78,7 +78,7 @@ def condition_match?(attribute, value)
 
     def hash_condition_match?(attribute, value)
       if attribute.is_a?(Array) || (defined?(ActiveRecord) && attribute.is_a?(ActiveRecord::Relation))
-        attribute.any? { |element| matches_conditions_hash?(element, value) }
+        attribute.to_a.any? { |element| matches_conditions_hash?(element, value) }
```
ghiculescu, line 81: @Liberatys if I remove the `.to_a` here, all tests still pass locally for me.

Do you know if I am missing something? 

I've tried testing against AR 5.2 and 6.0 on sqlite and postgres.

Liberatys: @ghiculescu the tests are mostly written to ensure / prove the functionality. The issue prominently arose in the setting of the example provided in the original issue. 

The discussion in the ticket outlined the issue and this change was made to address the behavior that was described in the ticket :)

ghiculescu: Got it. My concern is that if the tests pass even with the fix removed, then they aren't preventing us from regressing again.

I'm doing some refactoring of this code as part of https://github.com/CanCanCommunity/cancancan/pull/653 and I don't want to accidentally break it.

Do you think there is a test missing that I can include in that PR to ensure I don't regress?

PR #653, opened 2020-10-13
Support using a nil relation as a condition

ghiculescu:
I want to define an ability check that only applies if a `has_many` relation returns an empty array. The use case is I want to only allow deleting a record if it has no associated records for a given association.

In ActiveRecord you can do this with syntax like this:

> Article.joins(:comments).where(comments: { id: nil })

This generates the SQL:

> SELECT "articles".* FROM "articles" INNER JOIN "comments" ON "comments"."article_id" = "articles"."id" WHERE "comments"."id" IS NULL

Which returns an empty array.

This PR makes cancan support the same style of query. So you can do:

> @ability.can :destroy, Article, comments: { id: nil }

This will allow you to `:destroy` an `Article` that has no comments, but not an article that has comments.


spec/cancan/model_adapters/active_record_adapter_spec.rb
```diff
@@ -443,6 +443,173 @@ class User < ActiveRecord::Base
...
+    a1 = Article.create!
+    a2 = Article.create!
+    a2.comments = [Comment.create!(spam: true)]
+    a3 = Article.create!
+    a3.comments = [Comment.create!(spam: false)]
+
+    # if we are checking for `id: nil` and any other criteria, we should never return any Article.
+    # either the Article has Comments, which means `id: nil` fails.
+    # or the Article has no Comments, which means `spam: false` fails.
+    @ability.can :read, Article, comments: { id: nil, spam: false }
```
coorasse, line 548: does this work also if it's defined separately?
```
@ability.can :read, Article, comments: { id: nil }
@ability.can :read, Article, comments: { spam: false }
```

spec/cancan/model_adapters/active_record_adapter_spec.rb
```diff
@@ -443,6 +443,173 @@ class User < ActiveRecord::Base
...
+      expect(@ability.model_adapter(Article, :read)).to generate_sql(%(
+  SELECT "articles".*
+  FROM "articles"
+  WHERE "articles"."id" IN (SELECT "articles"."id" FROM "articles"
+    LEFT OUTER JOIN "comments" ON "comments"."article_id" = "articles"."id"
+    WHERE "comments"."id" IS NULL AND "comments"."spam" IS NULL)))
+    end
+  end
+
+  it 'doesn\'t allow a nil to be used as a condition for a has_many alongside other attributes' do
```
coorasse, line 510: I don't think that it does not allow it. it just does not return any result. right?

spec/cancan/model_adapters/active_record_adapter_spec.rb
```diff
@@ -443,6 +443,173 @@ class User < ActiveRecord::Base
...
+    @ability.can :read, Article, comments: { id: nil, spam: false }
+
+    expect(@ability.can?(:read, a1)).to eq(false)
+    expect(@ability.can?(:read, a2)).to eq(false)
+    expect(@ability.can?(:read, a3)).to eq(false)
+
+    expect(Article.accessible_by(@ability)).to eq([])
+
+    if CanCan::ModelAdapters::ActiveRecordAdapter.version_greater_or_equal?('5.0.0')
+      expect(@ability.model_adapter(Article, :read)).to generate_sql(%(
```
coorasse, line 557: why does it even try? 😄

coorasse (commented):
I have one or two questions and maybe a request for an additional test. good job!

spec/cancan/model_adapters/active_record_adapter_spec.rb
ghiculescu: Haha, good question. I tried to avoid adding too much logic specific to this scenario, and mostly rely on how the Rails querying interface already works. In this case, we are generating a query that can never be true, but Rails hasn't necessarily understood that or optimised it out.

ghiculescu: I suppose I could try and refactor it to return `Article.none` in this case, but it feels like it could add a fair bit more maintenance load.

spec/cancan/model_adapters/active_record_adapter_spec.rb
ghiculescu: It would be an [or statement](https://github.com/CanCanCommunity/cancancan/blob/develop/docs/Defining-Abilities.md#combining-abilities) in that case. I will add a test.

ghiculescu: https://github.com/CanCanCommunity/cancancan/pull/653/commits/a41c346d80b65a3c367c5f92537c8c969ca6e13e

spec/cancan/model_adapters/active_record_adapter_spec.rb
ghiculescu: https://github.com/CanCanCommunity/cancancan/pull/653/commits/6ebf456a98ebf7f89009156438ca394f480b2ee4

spec/cancan/model_adapters/active_record_adapter_spec.rb
```diff
@@ -443,6 +443,199 @@ class User < ActiveRecord::Base
...
+  WHERE 1=0))
+    end
+  end
+
+  it 'allows a nil to be used as a condition for a has_many - with join' do
+    a1 = Article.create!
+    a2 = Article.create!
+    a2.comments = [Comment.create!]
+
+    @ability.can :read, Article, comments: { id: nil }
```
coorasse, line 471: would `can :read, Article, comments: nil` work as well? I'd expect so from 3.2.0.

And what about `can :read, Article, comments: []`? Maybe I'd expect this to work as well as `can :read, Article, comments: { id: [] }`

ghiculescu: Not currently. The cancan conditions need to match conditions you'd pass into AR#where. Both of your examples wouldn't work as regular queries.

```ruby
(byebug) Article.left_outer_joins(:comments).where(comments: nil)
#<ActiveRecord::Relation []>
(byebug) Article.left_outer_joins(:comments).where(comments: nil).to_sql
"SELECT \"articles\".* FROM \"articles\" LEFT OUTER JOIN \"comments\" ON \"comments\".\"article_id\" = \"articles\".\"id\" WHERE \"articles\".\"id\" IS NULL" # note how it looks for article.ids IS NULL, not comments.id
```

```ruby
(byebug) Article.left_outer_joins(:comments).where(comments: [])
#<ActiveRecord::Relation []>
(byebug) Article.left_outer_joins(:comments).where(comments: []).to_sql
"SELECT \"articles\".* FROM \"articles\" LEFT OUTER JOIN \"comments\" ON \"comments\".\"article_id\" = \"articles\".\"id\" WHERE 1=0"
```

Whereas with my example:

```ruby
(byebug) Article.left_outer_joins(:comments).where(comments: { id: nil })
#<ActiveRecord::Relation [#<Article id: 1, name: nil, created_at: "2020-12-14 00:50:15", updated_at: "2020-12-14 00:50:15", published: nil, secret: nil, priority: nil, category_id: nil, project_id: nil, user_id: nil>]>
(byebug) Article.left_outer_joins(:comments).where(comments: { id: nil }).to_sql
"SELECT \"articles\".* FROM \"articles\" LEFT OUTER JOIN \"comments\" ON \"comments\".\"article_id\" = \"articles\".\"id\" WHERE \"comments\".\"id\" IS NULL"
```

ghiculescu: I'm open to trying to support more use cases, but I think it'd add a bit more complexity as we'd need to translate specific values (`nils`, `[]`, etc) into the corresponding AR query statements.

ghiculescu:
@coorasse anything more you need from me to get this merged?

coorasse:
I believe this is ready to be merged. Could you squash and rebase, please?

ghiculescu:
I'm just gonna run our CI suite against this now that a few changes upstream have been merged in.

coorasse:
Sure! Just let me know, but I believe is good to go 👍

PR #655, opened 2020-10-13
Add option for `accessible_by` querying strategy

ghiculescu:
Inspired by [this comment](https://github.com/CanCanCommunity/cancancan/pull/619#issuecomment-707683505), let users choose the querying strategy cancan will use when calling `accessible_by`.

You can choose `:subquery` (the default since https://github.com/CanCanCommunity/cancancan/pull/619), or `:left_joins` (how it worked before https://github.com/CanCanCommunity/cancancan/pull/619).

Most tests run exactly the same on both strategies. Where they differ I have split into two separate test contexts.

I have deliberately not added this to the README, because out of the box I don't think you should change this. I would like to add a mention to https://github.com/CanCanCommunity/cancancan/wiki/Fetching-Records if this is merged (I don't have edit access to that page).


ghiculescu:
@coorasse let me know your thoughts

madmax:
@ghiculescu I think it will be god if we can decide by ability when using subquery and when left_joins not only globaly. Some queries will be better with joins some with subquery.

ghiculescu:
I looked into doing it that way @madmax but it added a fair bit more complexity. I'm keen to start with this, because it allows everyone to match pre-https://github.com/CanCanCommunity/cancancan/pull/619 behavior. We can always add more options later.

coorasse:
I love the effort you put in this PR! There is a lot of good work! I have to agree with @madmax . I strongly believe that this option could/should be also on a "per call" level. Because most of the times you might want to change it only in one specific place. Nevertheless: I am also in favour of a global configuration, so I am keen in merging after a review.

lib/cancan/model_adapters/active_record_4_adapter.rb
```diff
@@ -34,10 +34,15 @@ def matches_condition?(subject, name, value)
...
-        relation
+      def build_joins_relation(relation, *_where_conditions)
+        case CanCan.accessible_by_strategy
+        when :subquery
+          # subquery mode doesn't work with Rails 4.x
+          relation.includes(joins).references(joins)
+
+        when :left_join
+          relation.includes(joins).references(joins)
+        end
```
coorasse, line 45: I guess I am missing something here 😅  why does it do the same?

coorasse (commented):
I really like your work! There's only one question that I have

coorasse:
I love the effort you put in this PR! There is a lot of good work! I have to agree with @madmax . I strongly believe that this option could/should be also on a "per call" level. Because most of the times you might want to change it only in one specific place. Nevertheless: I am also in favour of a global configuration, so I am keen in merging after a review.

lib/cancan/model_adapters/active_record_4_adapter.rb
ghiculescu: I'll tidy this up a bit. The config option really only works with AR5+.

ghiculescu:
I thought of a way of making this work per-query without cluttering up the cancan API a lot.

```ruby
CanCan.with_accessible_by_strategy(:left_joins) do
  User.accessible_by(current_ability)
end
```

Thoughts? Should be relatively easy to implement.

coorasse:
What about opening a discussion regarding this point? I just enabled the tab in the project. We can start from your proposal @ghiculescu . So we can proceed with this PR and merge it in the meantime and proceed elsewhere to decide how we want to implement the per-query API.

lib/cancan/config.rb
```diff
@@ -0,0 +1,49 @@
...
+      raise ArgumentError, "accessible_by_strategy must be one of #{valid_accessible_by_strategies.join(', ')}"
+    end
+
+    if value == :subquery && CanCan::ModelAdapters::ActiveRecordAdapter.version_lower?('5.0.0')
+      raise ArgumentError, 'accessible_by_strategy = :subquery requires ActiveRecord 5 or newer'
+    end
+
+    @accessible_by_strategy = value
+  end
+end
```
coorasse, line 49: I like this. very descriptive

coorasse (approved):
Maybe an update to the docs will be necessary, but it can be done in a separate PR. Thanks!

ghiculescu:
@coorasse happy to add docs, any suggestions on where? I was thinking https://github.com/CanCanCommunity/cancancan/blob/develop/docs/Defining-Abilities:-Best-Practices.md but I'm not sure any of the current docs feels right.

ps. I'm not sure what https://github.com/CanCanCommunity/cancancan/blob/develop/docs/mvc--deficiencies.md is but feel like it shouldn't be there.

coorasse:
Thanks for finding the irrelevant doc page, I removed it. The whole docs would need a bit of re-organisation. I'd definitely need some help there. 
Probably here? https://github.com/CanCanCommunity/cancancan/blob/develop/docs/Fetching-Records.md

PR #656, opened 2020-10-24
Update Wiki Links

lukasbischof:
The updated links in the readme lead to a redirect form wiki to a document in `docs/` folder.
Also for the `docs/Checking-Abilities.md`, it appears that those should be links. 
Anyway, in general it's a bit confusing what has more precedence - the docs folder or the wiki? Since some links lead to wiki and some lay in docs folder and some pages in docs are just copies of the ones in the wiki

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ of models automatically and reduce duplicated code.
 <br/>
 <br/>
 
-Do you want to sponsor CanCanCan and show your logo here? 
+Do you want to sponsor CanCanCan and show your logo here?
 Check our [Sponsors Page](https://github.com/sponsors/coorasse).
 
 ## Installation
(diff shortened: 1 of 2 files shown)


BKSpurgeon:
@lukasbischof 

The docs are now the source of truth and the wiki is now obsolete, and can be dispensed with. Alas, I don't have the permissions for it.

README.md
```diff
@@ -242,11 +242,11 @@ See [Ensure Authorization](https://github.com/CanCanCommunity/cancancan/wiki/Ens
 
 ## Wiki Docs
 
-* [Defining Abilities](https://github.com/CanCanCommunity/cancancan/wiki/Defining-Abilities)
+* [Defining Abilities](https://github.com/CanCanCommunity/cancancan/blob/develop/docs/Defining-Abilities.md)
```
coorasse, line 245: could you please link to the current branch? `(./docs/Defining-Abilities.md)` should do

coorasse (commented):
Thanks for the contribution @lukasbischof ! 🙏 
Could you please link using a relative path? So that each branch would point to the correct path

lukasbischof:
I updated the PR to include:
* `docs/` as the source of truth
* Relative paths to docs folder where applicable

coorasse (approved):

Issue #659, opened 2020-11-05
Use of subject.subclasses in #alternative_subjects problematic in big projects


23tux:
With this PR https://github.com/CanCanCommunity/cancancan/pull/649 a change was introduced, that breaks big projects:

https://github.com/CanCanCommunity/cancancan/blob/695d9b274075ab2408222e2e6e8b9e2867c81744/lib/cancan/ability.rb#L306

In our Rails app, we use `can?(:admin, Object)` to define some admin-only behaviour. The problem is

```
Benchmark.ms { puts Object.subclasses.count }
2994
=> 48140.53920000151
```

So it takes 48s to gather all the subclasses of `Object`. Can someone explain, WHY `subject.subclasses` is used here? I think it has something to do with the STI feature, but I guess this could be really problematic in big applications.

coorasse:
cc @Liberatys

23tux:
@coorasse I'm not too deep into the STI feature, but that sounds promising!

In the mean time, I have introduced a dummy object to check against

```
class AdminObject
end

can?(:admin, AdminObject)
```

Liberatys:
@23tux a fix is already in the making. 
https://github.com/CanCanCommunity/cancancan/pull/660
This should fix the issue for Object check

coorasse:
The PR should fix the problem. Allow me to say also that I never saw using `cancancan` this way. I'd be interested in knowing why you do that 😄

23tux:
@coorasse awesome, thank you!

That came from the Spree Commerce framework, where every action in the admin panel is tested against cancancan. And when you don't have explicitly defined a resource or model, it falls back to `Object`, to only allow admins access to those sites. Therefore you get calls like `can?(:admin, Object)` as a way of telling "Only the admin is allowed to do that".

PR #660, opened 2020-11-05
Feature/fix subclass check

Liberatys:
This addresses an issue with bigger projects, where classes have a big inheritance structure and thus the collection of subclasses is taking longer than is feasible. 

Introduce a guard to check if given element is inheriting from ActiveRecord::Base. Otherwise subclasses will not be collected because they are only needed in the process of creating sti connections.

The check with "<" should work, as you can not instantiate ActiveRecord::Base. Thus <= is not needed.

----

Adresses #659

cc @coorasse


coorasse (approved):
Thanks!

PR #673, opened 2020-12-13
Check for ActiveRecord in subject

### LINKED ISSUE #672: CanCanCan 3.2 requires ActiveRecord in ability alternative_subjects
In ability is a reference to ActiveRecord on line 305 in alternative_subjects, too. This causes issues if you don't use AR, but e.g. Mongoid

Liberatys:
Checks for ActiveRecord to fix #672 

Also address #676

### DIFF
--- a/lib/cancan/ability.rb
+++ b/lib/cancan/ability.rb
@@ -302,7 +302,7 @@ module CanCan
 
     def alternative_subjects(subject)
       subject = subject.class unless subject.is_a?(Module)
-      if subject.respond_to?(:subclasses) && subject < ActiveRecord::Base
+      if subject.respond_to?(:subclasses) && defined?(ActiveRecord::Base) && subject < ActiveRecord::Base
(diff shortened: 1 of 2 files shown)


lib/cancan/ability.rb
```diff
@@ -302,6 +302,9 @@ def extract_subjects(subject)
 
     def alternative_subjects(subject)
       subject = subject.class unless subject.is_a?(Module)
+      unless defined? ActiveRecord
+        return [:all, *subject.ancestors, subject.class.to_s]
+      end
       if subject.respond_to?(:subclasses) && subject < ActiveRecord::Base
```
coorasse, line 308: can't we change this if to `subject.respond_to?(:subclasses) && defined?(ActiveRecord::Base) && subject < ActiveRecord::Base` and avoid the code above?

coorasse (approved):

prcongithub:
Is new tag pushed for this?

PR #675, opened 2020-12-14
Support modifying the `accessible_by` querying strategy on a per-query basis

ghiculescu:
See discussion at https://github.com/CanCanCommunity/cancancan/discussions/668


coorasse (approved):
Thank you!

NikosVlagoidis:
I came across that and noticed that the implementation isn't thread-safe and I would be happy to provide a patch with some guidance. My initial thought was to use [thread_mattr_accessor](https://apidock.com/rails/v5.2.3/Module/thread_mattr_accessor) from active_support but this isn't a repo dependency.

Would going with a `Thread.current[]` variable be reasonable?

ghiculescu:
That sounds like a good idea

PR #682, opened 2020-12-30
Use bundler-cache: true to automatically cache installed gems

eregon:


eregon:
```
Run bundle exec rubocop
Error: Unsupported Ruby version 2.1 found in `TargetRubyVersion` parameter (in vendor/bundle/ruby/2.4.0/gems/rainbow-3.0.0/.rubocop.yml). 2.1-compatible analysis was dropped after version 0.58.
Supported versions: 2.2, 2.3, 2.4, 2.5, 2.6
Error: Process completed with exit code 2.
```
That's RuboCop looking at files it should not, will try to tell it not to do that.

eregon:
@coorasse Passing now :)

I guess TravisCI could be removed since GHA works fine?

artplan1:
hi @eregon, I just checked issue with `exclude` (disscused in https://github.com/ruby/setup-ruby/issues/131) and it seems it works correctly by default.

_But quotes in gemfiles should be fixed (or cop should be disabled for gemfiles) :)_

failed build because of quotes: https://github.com/artplan1/cancancan/runs/1626274482?check_suite_focus=true#step:6:13
green build with fixes: https://github.com/artplan1/cancancan/runs/1626294534?check_suite_focus=true#step:6:13
changes: https://github.com/artplan1/cancancan/pull/1/files

eregon:
> But quotes in gemfiles should be fixed (or cop should be disabled for gemfiles) :)

I assume that's not an option, because these gemfiles are generated by Appraisal.
Although, there might be a newer version of Appraisal or some way to configure it to use single quotes.
Still, I'd like to keep it simple in this PR :)

eregon:
@coorasse Could you merge this?
It also excludes a failing combination of TruffleRuby + Rails 4 (cannot work, TruffleRuby now targets Ruby 2.7 which doesn't work on Rails 4).

coorasse (approved):
Thanks!

Issue #677, opened 2020-12-18
STI support in 3.2.0 is a breaking change


Fryie:
### Steps to reproduce

So, here is some fun change in behaviour in the 3.2.0 release, related to the STI support.

The following tests fail in 3.2.0, but they pass in the previous 3.1.0 version: https://gist.github.com/Fryie/46b759ca19efa28e1dfdb41a54742e1b

Basically, calling index on some resource may lead to an authorization failure if access to a subclass of that resource is disallowed. Changing the order of the two ability rules also affects the code in a somewhat unintuitive(?) way: the previously failing assertion passes, but the one that used to pass doesn't anymore. (This seems due to the fact that matching rules are selected in reverse order).

### Expected behavior

I'm not sure what the *expected* behaviour should be here. Inheritance generally makes reasoning about something like this harder: should I be able to index all `parents` even if some of them might actually be `children`, which I'm not technically able to access? Maybe not (but then again, what if my controller automatically filters the children out?)? But probably this change is also unexpected in a minor release. At least a gotcha/workaround could be added somewhere in the documentation.

The problem is unfortunately compounded in my case by the fact that [ActiveAdmin doesn't distinguish between index and show](https://github.com/activeadmin/activeadmin/issues/4704). So saying something like "the user should be able to list all parents (including, or excluding children), but the show action should be forbidden for children" doesn't actually seem to be possible. As far as I am aware, this used to work in 3.1.0.

### Actual behavior
See above

### System configuration
**Rails version**:
see gist

**Ruby version**:
ruby 2.7.1p83 (2020-03-31 revision a0c7c23c9c) [x86_64-darwin19]

**CanCanCan version**
see gist

coorasse:
The test on line https://gist.github.com/Fryie/46b759ca19efa28e1dfdb41a54742e1b#file-cancan_bug-rb-L62 should definitely return `true` in my opinion, so we have a regression.
`can :index, ModelClass` should return `true` since you can `index` the class itself. I don't think is possible to use STI here in a smart way.

@Liberatys what is your opinion on this issue?

tagliala:
Hi,

I'm experiencing a similar regression with breaking change when permissions is only assigned to a subclass, propagating to the parent class

```rb
class Motorbike < ApplicationRecord
end

class Suzuki < Motorbike
end

# frozen_string_literal: true

class Ability
  include CanCan::Ability

  def initialize(user = nil)
    can :manage, Suzuki
  end
end
```

[...]

PR #689, opened 2021-01-24
Implement detection for sti for including subclasses in check

Liberatys:
* Remove subclass rule inclusion for non sti classes
* Start of implementing more sti detection utilities which are going to be used in a rewrite (coming)

@coorasse how do you want to handle this ? It seems like STI will have some issues with the current implementation. I'll probably have to add conditional loading at 3 places to fully comply with the expected behaviour.


lib/cancan/class_matcher.rb
```diff
@@ -12,6 +14,8 @@ def self.matches_subject_class?(subjects, subject)
   def self.matching_class_check(subject, sub, has_subclasses)
     matches = matches_class_or_is_related(subject, sub)
     if has_subclasses
+      return matches unless StiDetector.sti_class?(sub)
```
Liberatys, line 17: Only include the subclasses in the check if the model is an sti class

lib/cancan/sti_detector.rb
```diff
@@ -0,0 +1,10 @@
+class StiDetector
+  def self.sti_class?(subject)
```
Liberatys, line 2: A general detection for STI classes

spec/cancan/model_adapters/active_record_adapter_spec.rb
```diff
@@ -909,6 +909,37 @@ class JsonTransaction < ActiveRecord::Base
     end
   end
 
+  context 'with rule application to subclass for non sti class' do
```
Liberatys, line 912: Check that rules are no longer applied to parents for non sti

spec/cancan/model_adapters/active_record_adapter_spec.rb
```diff
@@ -909,6 +909,37 @@ class JsonTransaction < ActiveRecord::Base
...
+      end
+
+      class Parent < ActiveRecord::Base
+      end
+
+      class Child < Parent
+      end
+    end
+
+    it 'rules are not effecting parent class' do
```
phallstrom, line 933: I think we also need the test where can'ing the child doesn't automatically can the parent to satisfy this scenario, yeah? https://github.com/CanCanCommunity/cancancan/issues/677#issuecomment-753387317

Something like...

```
u1 = User.create!(name: 'pippo')
ability = Ability.new(u1)
ability.can :manage, Child
expect(ability).to be_able_to(:index, Child)
expect(ability).not_to be_able_to(:index, Parent)
```

winstonwolff:
Thank you for working on this. We've tried this code and it seems to work—our test cases pass and it seems to block a parent class of resources as we expected. If we see any problems with it, we'll report back.

swelther:
Any news on this? 🙏 This issue prevents us from updating to cancancan 3.2+ :(

lib/cancan/sti_detector.rb
coorasse: Love its isolation

coorasse (commented):
@Liberatys do we need to check further scenarios here, like @phallstrom suggested, or are they covered already somewhere else?

Liberatys:
@coorasse I now also added the spec that covers the example from #677. CI is blue/green ^^

coorasse (approved):

PR #691, opened 2021-01-26
Accessible by strategy: Exists subquery

kaspernj:
The new subquery strategy is great, but doing a "WHERE id IN (...)" can be slow, because Postgres will resolve the ID's on the right first, so if you have a big table that is going to take a lot of juice.

"WHERE EXISTS (...)" should be a lot faster, because it will already know the related ID's and only sub query those instead of resolving them first through the given sub query.

There is a problem however. It's difficult to alias the table properly in ActiveRecord. I have gotten around that by doing two exist-queries just to make it possible to reference the original table in the exists-subquery. It's still much faster than doing a "WHERE id IN (...)" however and it solves my own performance issues.

When I try to run specs with the following command I get an error: `uninitialized constant CanCan::ModelAdapters::ActiveRecord5Adapter`

Do I need to do something special to get the specs up and running?


ghiculescu:
> Do I need to do something special to get the specs up and running?

Rebase against master, now that Travis has been replaced by Github Actions https://github.com/CanCanCommunity/cancancan/pull/707

kaspernj:
> Rebase against master, now that Travis has been replaced by Github Actions #707

I still get the same error when trying to run the specs locally. I have tried running `bundle exec rake` and `bundle exec rspec` but both result in the same error unfortunately :'-(

lib/cancan/model_adapters/active_record_5_adapter.rb
```diff
@@ -21,17 +21,57 @@ def self.matches_condition?(subject, name, value)
...
+
       def build_joins_relation(relation, *where_conditions)
         case CanCan.accessible_by_strategy
         when :subquery
-          inner = @model_class.unscoped do
-            @model_class.left_joins(joins).where(*where_conditions)
-          end
-          @model_class.where(@model_class.primary_key => inner)
-
+          build_joins_relation_subquery(where_conditions)
```
kaspernj, line 30: This refactor was needed to deal with Rubocop complexity offences.

lib/cancan/model_adapters/active_record_5_adapter.rb
```diff
@@ -21,17 +21,57 @@ def self.matches_condition?(subject, name, value)
...
+          @model_class.left_joins(joins).where(*where_conditions)
         end
+        @model_class.where(@model_class.primary_key => inner)
+      end
+
+      def build_joins_relation_double_exist_subquery(where_conditions)
+        @model_class.where("EXISTS (#{double_exists_query_sql(where_conditions)})")
+      end
+
+      def double_exists_inner_query(where_conditions)
```
kaspernj, line 49: I am not sure the code has gotten less abstract having to use this many helper methods, but I couldn't otherwise find a way around the Rubocop complexity offences :'-(

coorasse: I honestly might get rid of rubocop and switch to standard soon :)

lib/cancan/model_adapters/active_record_5_adapter.rb
coorasse: we should probably, at some point, extract these strategies

coorasse:
I really like this PR, and I really like the fact that we start working on different strategies (SQL) to fetch the records. cancancan cannot always know what's the best strategy, so the only way we have is to allow users to use custom strategies. 

Being said that. I'd like to include this new strategy. I want to first pull the branch and see if I can add more tests for more cases

kaspernj:
@coorasse That sounds great. We cannot live without this strategy in the project I am working on, which involves a lot of really big tables with tons of data :-)

lib/cancan/model_adapters/active_record_5_adapter.rb
kaspernj: @coorasse I have extracted all strategies as separate classes (also to solve Rubocop complexity issues). What do you think?

lib/cancan/model_adapters/active_record_5_adapter.rb
kaspernj: Personally I feel that Rubocop has forced me to make the code more abstract than it had to be here :-)

kaspernj:
I have refactored to two different strategies:

*Joined alias exists subquery*
Joins the same table with an alias, which is then used to make an exists query, which is faster than "WHERE id IN (...)". You will need to use the full table name for columns, since the alias will have the same columns - but you should always do that in ActiveRecord anyway ;-) 

* Joined alias each rule as exists *
Joins the same table with an alias, iterates through all relevant rules, and uses them as separate or-exists queries. This can be very useful to avoid ActiveRecord messing up the join names, if the same tables are used in multiple other different rules, which is something that has plagued us a lot. The drawback is that it is a lot slower on large tables, but it might be your only way out, if ActiveRecord decides to mess you up, which is currently the case for us in Rails 6.1 (but works in 6.0.4 but not in 6.0.3). Here you must also use the table names for columns to not mess with the alias.

lib/cancan/model_adapters/active_record_5_adapter.rb
```diff
@@ -22,16 +22,13 @@ def self.matches_condition?(subject, name, value)
...
-          end
-          @model_class.where(@model_class.primary_key => inner)
+        strategy_class.new(adapter: self, relation: relation, where_conditions: where_conditions).execute!
+      end
 
-        when :left_join
-          relation.left_joins(joins).distinct
-        end
+      def strategy_class
+        require_relative "strategies/#{CanCan.accessible_by_strategy}"
```
xjunior, line 29: Why not just require them at the top of the file?

kaspernj: That would require code that the user in most cases would never use. I assume that most users are just going to set a single strategy and use that for everything :-)

coorasse:
Thanks! I went through the PR and looks solid. I'll probably extract some parts of it and merge them separately.

coorasse:
@kaspernj regarding the tests, you can run them using appraisal (see README). For example:

```
DB='sqlite' bundle exec appraisal rake
```

coorasse:
I am now looking at the first of the two stretegies, and I don't understand why you need to join the table with itaself....:🤔 

https://github.com/CanCanCommunity/cancancan/pull/691/files#diff-cf63e8e34aa3f5976e07653a98befdb9ae3cabef3996551d81352afdaa9e0f71R9

kaspernj:
@coorasse The short version is to not dynamically join tables directly in the query, which can cause issues when using other libraries that work with dynamic queries like Ransack (and others) :-)

Here is an example.

1. Multiple different roles that has different types of access.
2. Based on which role the current user has a number of tables are joined.
3. Same table is joined by different queries.
4. Because the same table is joined multiple times the aliases changes based on which role the user has.

To get around this a sub query is used where CanCan joins the table. Manually joined tables (or tables joined through other libraries like Ransack) are unaffected by which role the user has.

This can also result in bugs, because joining tables with aliases can mess up in ActiveRecord. I have had multiple issues about this in both Rails and Ransack already. By joining the tables from CanCan in a sub query, we can minimise those issues greatly, which directly fixed a number of issues in two projects I was working :-)

kaspernj:
@coorasse So I see the two new strategies as tools to get around those issues. Maybe you wouldn't use these strategies by default, but they would be options to get around the mentioned issues if you run into them like I did :-)

kaspernj:
> I am now looking at the first of the two stretegies, and I don't understand why you need to join the table with itaself....:🤔
> 
> https://github.com/CanCanCommunity/cancancan/pull/691/files#diff-cf63e8e34aa3f5976e07653a98befdb9ae3cabef3996551d81352afdaa9e0f71R9

@coorasse In order to do a sub query, I cannot use the original table name, so I have to alias the same table to use its ID in the sub query. If you know a better way of doing this then please enlighten me? :-)

coorasse:
I tried a lot of different things, but I am not able to generate an alias 😞 . I think your solution is the only working one, although I still really don't like that it performs a join that is not necessary. Pity. I'll think this through a bit more today, but if I don't find an alternative, I'd rather have this merged than not. So expect to have this available in the next release of cancancan anyway.

kaspernj:
@coorasse I feel the exact same way about the alias, and I have also sunk a considerable amount of time into trying to get around it :'-(

kaspernj:
The performance gains of this has been huge however :-)

PR #583, opened 2019-04-13
Fix method reference block regression

coorasse:
Fixes #582

PR #692, opened 2021-02-09
Preserve nil block arguments

remofritzsche:
Since 3.1.0 (#583), `nil` values are automatically stripped from extra block arguments. This is cumbersome when you want to pass `nil` values in some cases.

This PR aims to fix that while preserving the fix in #583.

I'm not quite sure though whether this fix makes sense, as I do not fully grasp the concept of the `attribute` parameter which, to my understanding, was the reason to put the `compact` there in the first place.

Please let me know what you think. Many thanks.


coorasse:
I took care of the merge conflicts, and merged it. Thanks!

PR #695, opened 2021-02-24
Use `relation.and` instead of `relation.merge` whenever possible.

albb0920:
Currently cancancan uses `relation.merge` to apply scope in ability rule to the scope of `load_resource through`.
However, since `relation.merge` intends to mimics Hash#merge,
(i.e. later condition on the same column replaces the previous one), 
this causes the `load_resource through` become ineffective when the scope in ability also specifics conditions on the same column.

This PR adds model adapter for activerecord >= 6.1 and overrides `database_records` to use `relation.and` instead. It also tries to keep the query simple by only wrapping scopes to sub queries only when necessary, which relies on private ActiveRecord API `structurally_incompatible_values_for`

## Example: 
ability file:
```ruby
can :manage, Group

can :manage, Post, Post.where(group_id: user.groups) do |post|
  post.gorup.users.exists?(user.id)
end
```

controller:
```ruby
class PostsController < ApplicationController
  load_and_authorize_resource :group
  load_and_authorize_resource through: :group
  
  def index
    # @posts will be from all groups,
    # as the condition for Post#group_id setted on `load_and_authorize_resource through`
    # is replaced by the condition specified in ability rule during relation.merge 
  end
end
```

## References
For more information on the behavior of `relation.merge` and `relation.and` that was introduced in rails 6.1,
please refer to rails/rails#39250 & rails/rails#40944

### DIFF
--- a/lib/cancan.rb
+++ b/lib/cancan.rb
@@ -21,4 +21,5 @@ if defined? ActiveRecord
   require 'cancan/model_adapters/active_record_adapter'
   require 'cancan/model_adapters/active_record_4_adapter'
   require 'cancan/model_adapters/active_record_5_adapter'
+  require 'cancan/model_adapters/active_record_61_adapter'
 end
--- a/lib/cancan/model_adapters/active_record_5_adapter.rb
+++ b/lib/cancan/model_adapters/active_record_5_adapter.rb
@@ -6,7 +6,7 @@ module CanCan
       AbstractAdapter.inherited(self)
 
       def self.for_class?(model_class)
-        version_greater_or_equal?('5.0.0') && model_class <= ActiveRecord::Base
+        version_greater_or_equal?('5.0.0') && version_lower?('6.1.0') && model_class <= ActiveRecord::Base
       end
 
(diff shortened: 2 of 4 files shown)


coorasse:
I'd like to tackle this by the next release that focuses on nested associations. Would it be possible for you to add the test for this case and rebase on the latest `develop`? If you don't have time, I'll take over. Thanks for your feedback and work

albb0920:
@coorasse I won't have a lot of free time until mid April, plus I'm not very familiar with the test suite of this project. Please take this over, thanks!

kwent:
Would love to see this one go through :)

s-mage:
Hey @coorasse thanks for a wonderful lib you're working on! Could I bring your attention to this PR please? It's already explained in the description but yeah, we have a very similar problem and the test that is missing would be 

```
can :manage, Post, Post.where(group_id: user.groups)
load_resource :posts, through: :group 
# should return effectively Post.where(group_id: user.groups).where(group_id: id_from_the_path)
# now returns Post.where(group_id: user.groups)
```

PR #698, opened 2021-03-14
Fix name in CI exclude

eregon:
I noticed I forgot to change the name for `ruby:` in the exclude line added in https://github.com/CanCanCommunity/cancancan/pull/682.
Now it should be good :)


coorasse:
Tackled in #707 . Thanks!

PR #705, opened 2021-04-26
[WIP] Introduce load_and_authorize_resource! alias

schmijos:
In code reviews I find myself often asking wether usage of `load_and_authorize_resource` was safe.
Then I'll look it up in the source code and see that indeed `authorize!` is being used.

I'd like to suggest that code writers from now on per default use `load_and_authorize_resource!` 
to transport the following message to the reviewer: "this is safe, no worries".

### DIFF
--- a/lib/cancan/controller_additions.rb
+++ b/lib/cancan/controller_additions.rb
@@ -15,6 +15,7 @@ module CanCan
       def load_and_authorize_resource(*args)
         cancan_resource_class.add_before_action(self, :load_and_authorize_resource, *args)
       end
+      alias load_and_authorize_resource! load_and_authorize_resource
 
       # Sets up a before filter which loads the model resource into an instance variable.
       # For example, given an ArticlesController it will load the current article into the @article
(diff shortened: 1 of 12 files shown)


lib/cancan/controller_additions.rb
```diff
@@ -15,6 +15,7 @@ module ClassMethods
       def load_and_authorize_resource(*args)
         cancan_resource_class.add_before_action(self, :load_and_authorize_resource, *args)
       end
+      alias_method :load_and_authorize_resource!, :load_and_authorize_resource
```
schmijos, line 18: Here I define the alias.

spec/cancan/controller_additions_spec.rb
```diff
@@ -49,6 +49,11 @@
     @controller_class.load_and_authorize_resource foo: :bar, prepend: true
   end
 
+  it 'load_and_authorize_resource is aliased by load_and_authorize_resource!' do
+    expect(@controller_class).to receive(:load_and_authorize_resource)
+    @controller_class.load_and_authorize_resource! foo: :bar
+  end
```
schmijos, line 55: And here I see if it can be used.

coorasse (commented):
Thank you @schmijos, I agree on promoting the BANG version as default: what about the method `authorize_resource`? Shall we alias this one as well?

I am also wondering about `load_resource`, but this method actually calls `find`, which indeed raises an exception but has no bang, so I am fine with it.

coorasse:
Please feel free to rebase and re-open.

PR #706, opened 2021-04-27
fix: use `sti_name` for normalizing STI type conditions

fsateler:
Rails itself uses this method. It can be used for example to have `Foo::Bar` only store `Bar` on the `type` column.

It is also (ab)used by the [active_type](https://github.com/makandra/active_type) gem to create derived-but-not-sti classes.


fsateler:
cc @Liberatys because this modifies their code.

xjunior (approved):

coorasse:
This has been covered in #689

PR #708, opened 2021-04-28
Add docs on securing nested resources

ghiculescu:
We recently were made aware of a potential security hole in our usage of `load_and_authorize_resource :through`.

Now that we have patched it, I thought some docs on how to prevent it would be helpful.


coorasse (approved):
Nice. Thanks!

PR #709, opened 2021-04-29
Goboony sponsorship

abuisman:
Hi everyone, 

I am pleased to announce that Goboony is now a CanCanCan sponsor. In this way we hope to express our thanks to Alessandro and the rest of the contributors for what is one of the fundamental gems for building Rails applications.

Alessandro has been kind enough to allow us to post our logo with a link. We have been so impudent to make the link go to an opening for a Full stack Rails developer 😇. Obviously we hope to persuade developers to have a look at what we have to offer, if not, you could always have a look at renting a camper van or renting yours out 😉.

Personally I am very happy that my employer has been so receptive to my wishes to give back to the community for the projects that we use, so hopefully you will see our name in some more places as well.

Thanks again!


coorasse (approved):
Thanks for your contribution! ❤️

PR #717, opened 2021-06-28
Support combining scopes when loading records, use Arel instead of SQL

mamhoff:
This refactors the ActiveRecordAdapter to use Arel/ActiveRecord to merge conditions. The advantage is that we can finally merge scopes from different abilities! 

This only works with Rails versions 5.2+, so this PR would drop support for prior versions. Given that Rails 7 is about to come out, that's something at least worth considering. 

I think this PR makes the ActiveRecord Adapter quite a bit more readable, which should make future contributions easier. 

I know it's a big change. Thank you for your hard work on this gem.


lib/cancan/model_adapters/active_record_adapter.rb
```diff
@@ -3,68 +3,40 @@
...
 
       def database_records
-        if override_scope
-          @model_class.where(nil).merge(override_scope)
-        elsif @model_class.respond_to?(:where) && @model_class.respond_to?(:joins)
-          build_relation(conditions)
+        if @model_class.respond_to?(:where) && @model_class.respond_to?(:joins)
+          build_relation
         else
           @model_class.all(conditions: conditions, joins: joins)
```
xjunior, line 27: This looks like support for old AR

lib/cancan/model_adapters/active_record_adapter.rb
```diff
@@ -3,68 +3,40 @@
...
-        end
-      end
-
-      def extract_multiple_conditions(conditions_extractor, rules)
-        rules.reverse.inject(false_sql) do |sql, rule|
-          merge_conditions(sql, conditions_extractor.tableize_conditions(rule.conditions).dup, rule.base_behavior)
+        ConditionsNormalizer.normalize(@model_class, @compressed_rules)
+        # run the extractor on the reversed set of rules to fill the cache properly
+        @conditions_extractor = ConditionsExtractor.new(@model_class).tap do |extractor|
+          @compressed_rules.reverse_each { |rule| extractor.tableize_conditions(rule.conditions) }
```
xjunior, line 19: I wonder if these `reverse`'s are necessary (line 14, then re-reversed on line 19)

lib/cancan/model_additions.rb
```diff
@@ -22,7 +22,7 @@ module ClassMethods
       # Here only the articles which the user can update are returned.
       def accessible_by(ability, action = :index, strategy: CanCan.accessible_by_strategy)
         CanCan.with_accessible_by_strategy(strategy) do
-          ability.model_adapter(self, action).database_records
+          ability.model_adapter(all, action).database_records
```
xjunior, line 25: What if this is already being called from a scope?

```ruby
User.active.accessible_by(ability, :read)
```

I think that switching this to `all` causes `accessible_by` to return a new scope ignore all previous filters

lib/cancan/model_adapters/active_record_adapter.rb
mamhoff: They are, because the `RulesCompressor` has an implicit requirement on the ordering of the conditions, unfortunately.

lib/cancan/model_additions.rb
mamhoff: The opposite is true: current behavior is to ignore all previous filters, and this change will take them into account. 

Try in any project you have: 

`User.active.all.count` vs
`User.active.klass.all.count` - using `klass` (or `self` in the above code) will unscope the relation, whereas `all` will not.

xjunior: Gotcha, I thought `self` in that context was a AR relation.

23tux:
Is there any news on this?

coorasse:
I am ready to reconsider this when Rails 8 is out Septemeber this year. I will then drop support for Rails 5 and this PR can be simplified. Thanks for the amazing work you did here!

PR #722, opened 2021-08-02
New documentation

coorasse:
A whole refactoring of the current documentation.


lib/generators/cancan/ability/templates/ability.rb
```diff
@@ -26,7 +24,7 @@ def initialize(user)
     # objects.
     # For example, here the user can only update published articles.
     #
-    #   can :update, Article, :published => true
+    #   can :update, Article, published: true
     #
     # See the wiki for details:
     # https://github.com/CanCanCommunity/cancancan/wiki/Defining-Abilities
```
mdondorf-avarteq, line 30: Hi @coorasse meanwhile this URL is outdated, should probably be changed to:
https://github.com/CanCanCommunity/cancancan/blob/develop/docs/define_check_abilities.md

buncis:
this broke separate role model link in https://github.com/CanCanCommunity/cancancan/blob/develop/docs/role_based_authorization.md

PR #724, opened 2021-08-03
Avoid using break in RSpec match which uses define_method

eregon:
* As that causes an issue with TruffleRuby 21.2.0: https://github.com/oracle/truffleruby/issues/2418


coorasse (approved):
Thanks!

PR #739, opened 2021-10-07
Updated Devise.md as exception is changed

Antronin:
CanCan::Unathorized exception not exists anymore, CanCan::AccessDenied is used. Also added responses for different content types as a recommendation.


coorasse (approved):
Agree. Thanks!

PR #741, opened 2021-10-12
Include LICENSE file in gem

marvinthepa:
Make it easier for users of the gem to get to the full license text
including copyright notice.

### DIFF
--- a/cancancan.gemspec
+++ b/cancancan.gemspec
@@ -16,7 +16,7 @@ Gem::Specification.new do |s|
   s.platform    = Gem::Platform::RUBY
   s.license     = 'MIT'
 
-  s.files       = `git ls-files lib init.rb cancancan.gemspec`.split($INPUT_RECORD_SEPARATOR)
+  s.files       = `git ls-files lib init.rb cancancan.gemspec LICENSE`.split($INPUT_RECORD_SEPARATOR)
   s.require_paths = ['lib']
 
   s.required_ruby_version = '>= 2.2.0'


coorasse:
Thanks for your contribution. I am not really willing to distribute this file on each client since is not necessary to the library usage. The license is very simple and already specified in the gemspec: https://github.com/CanCanCommunity/cancancan/blob/develop/cancancan.gemspec#L17

No need for further details for something very simple. Thanks!

marvinthepa:
Hey @coorasse,

I totally understand that you want to keep the gem lean and mean.

But please bear with me for a minute and let me explain the use case for this.
Disclaimers:
a) I am not a lawyer and I might be (very) wrong.
b) you are the boss here, I am just kindly asking for the license to be included - still totally up to you.

So:
If we want to adhere to this sentence

```
The above copyright notice and this permission notice shall be
included in all copies or substantial portions of the Software.
```

we HAVE TO include the full license (including the copyright statement) with it, not just the fact that it is MIT licensed. I cannot easily get this from the gemspec, and neither is it easy to automate to get it from github - I could hardcode the location for cancancan (as the gemspec does not tell me where to get the sources), but that will not help me for other gems - and it might break as well, e.g. if someone else starts to maintain the gem elsewhere.

And if you think about it, is not the gem itself a "copy or substantial portion of the Software"? Should it not then contain the license, or do you expect anyone who wants to redistribute it (e.g. by offering a caching rubygems proxy) to unpack it, add the license themselves, and then pack it again? I do not know (again, not a lawyer), but you might even be bound by the license yourself (If you do not in addition have the "all rights reserved" copyright that the original author always has).

PR #747, opened 2021-10-24
replace the check abilities wiki link with new doc link.

younthu:
the wiki link in abilities.rb template is not valid any more, replace it with new md doc link.


coorasse (approved):
thanks!

PR #752, opened 2021-11-08
Add Ruby 3.0 for CI

AlexWayfer:
Avoid incompatible RoR versions.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,9 +14,19 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: ['2.4', '2.5', '2.6', '2.7', 'jruby', 'truffleruby']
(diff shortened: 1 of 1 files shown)


coorasse (approved):

AlexWayfer:
Why jobs were canceled? 😓

coorasse:
Ruby 3 incompatible with Rails 4 and 5 so we need to exclude them from the tests matrix. can you please take care of it and also rebase? Thanks! 🙏

AlexWayfer:
> Ruby 3 incompatible with Rails 4 and 5 so we need to exclude them from the tests matrix. can you please take care of it and also rebase? Thanks! pray

Yep, of course, just waited for CI runs to figure out specific failing combinations.

Liberatys:
@coorasse is this good to go? :D

coorasse (approved):

PR #754, opened 2021-11-10
Extract strategies in separate files

coorasse:
Co-authored-by: kaspernj


rocket-turtle:
I think it's a good idea to extract this code and merge it in smaller junks.
I have just a quick question to this MR.

Why do you add all this delegations to `strategies/base.rb, I don't see where they get used.

And there is a method called `quoted_aliased_table_name` and one called `quoted_table_name` but they are also in the delegation to `adapter`. Is that on purpose?

Also the method `quoted_primary_key` is defined in two different delegations.

Or am I wrong and the delegation part works different here.

kaspernj:
@rocket-turtle They were used in the other strategies that I made, which weren't included in this PR. So ideally they should only have been in the other PR probably  :-)

PR #755, opened 2021-11-11
Feature/742 remove dependency on active record

Liberatys:
Check for the presence of ActiveRecord before using it


Liberatys:
@coorasse This should solve the issue of ActiveRecord not being present

PR #756, opened 2021-11-11
Feature/716 handle calls to subclass

Liberatys:
Implement memo for subclass to prevent duplicate subclass calls


lib/cancan/subject_class_matcher.rb
```diff
@@ -2,23 +2,28 @@
...
       matching_class_check(subject, sub, has_subclasses)
     end
   end
 
-  def self.matching_class_check(subject, sub, has_subclasses)
+  def matching_class_check(subject, sub, has_subclasses)
     matches = matches_class_or_is_related(subject, sub)
     if has_subclasses
-      matches || subject.subclasses.include?(sub)
+      @subclasses[subject.class.name] ||= subject.subclasses
```
Liberatys, line 19: this could probably fail if classes are manipulated at runtime? 🙃

lib/cancan/rule.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 require_relative 'conditions_matcher.rb'
-require_relative 'class_matcher.rb'
+require_relative 'subject_class_matcher.rb'
```
Liberatys, line 4: ... now fits the class name

Liberatys:
@coorasse happy with this?

PR #757, opened 2021-11-12
Debug exists subquery

coorasse:
This PR is a draft to debug the behavior of an exists subquery implemented in #691 

Run the relevant test with:

```
DB='sqlite' bundle exec appraisal activerecord_6.0.0 rspec spec/cancan/model_adapters/accessible_by_has_many_through_spec.rb[1:3:2:1]
```

My debugging started because I wanted to understand why we need this JOIN: https://github.com/CanCanCommunity/cancancan/pull/691/files#diff-1ac9b926205ddc47154cac149a2a53d61c308efc15b59bfdea7b13ec6565fff1R63-R64

Apparently is because otherwise it is basically impossible to alias the tables in the inner exists subquery.

### DIFF
--- a/spec/cancan/model_adapters/accessible_by_has_many_through_spec.rb
+++ b/spec/cancan/model_adapters/accessible_by_has_many_through_spec.rb
@@ -89,7 +89,8 @@ RSpec.describe CanCan::ModelAdapters::ActiveRecord5Adapter do
       if CanCan::ModelAdapters::ActiveRecordAdapter.version_greater_or_equal?('5.0.0')
         describe 'selecting custom columns' do
           it 'extracts custom columns correctly' do
-            posts = Post.accessible_by(ability).where(published: true).select('title as mytitle')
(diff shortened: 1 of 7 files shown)


spec/cancan/model_adapters/accessible_by_has_many_through_spec.rb
```diff
@@ -89,7 +89,8 @@ class Editor < ActiveRecord::Base
       if CanCan::ModelAdapters::ActiveRecordAdapter.version_greater_or_equal?('5.0.0')
         describe 'selecting custom columns' do
           it 'extracts custom columns correctly' do
-            posts = Post.accessible_by(ability).where(published: true).select('title as mytitle')
+            posts = Post.where(title: 'a').accessible_by(ability).where(published: true).select('title as mytitle')
+            puts posts.to_sql
```
coorasse, line 93: check output to see the issue

PR #763, opened 2022-01-10
Separate Test and Lint

mishina2228:
This PR separates Test and Lint to increase the stability of CI.
I believe that lint doesn't need to be run on various Ruby versions, so RuboCop will be run only once.


coorasse (approved):
Thanks!

PR #764, opened 2022-01-10
Adds Ruby 3.0 and 3.1, ActiveRecord 7 to CI

petergoldstein:
Ruby 3.1 and Rails 7 were released at the tail end of 2021.  And Ruby 3.0 was released at the end of 2020.  These should be represented in the CI matrix.

In addition to the basic CI configuration changes there are a few additional changes in this PR:

1. Upgrading Rubocop to a '~> 1' version to allow loading across Ruby versions
2. Setting the TargetRubyVersion to 2.5 for Rubocop, as that's the lowest Ruby version supported by recent Rubocop versions.
3. Explicitly disabling a few cops that are potentially problematic for earlier versions of Ruby.
4. Adding an ActiveRecord 7.0 gemfile
5. Adding an autogenerated .rubocop_todo.yml


coorasse:
Thanks. I resolved merge conflicts, rebased and merged

Issue #661, opened 2020-11-11
Children / sub-actions accessible even if object is unauthorized


bars0um:
### Steps to reproduce

This seems like a serious bug, or possibly I'm unable to find the proper documentation to do this correctly.

I have a controller for a class Order that has many OrderItems.

If I setup my controller as follows to perform authorization on the loaded instance variable, the behavior is not as expected when touching child objects or performing custom actions.

The following is the setup I have:

```
class OrdersController < ApplicationController
  before_action :find_order
  load_and_authorize_resource
  load_and_authorize_resource :order_item, through: :order
...
..
def find_order
    @order = if params[:order_id]
               Order.find(params[:order_id])       
    end
```

I have a nested route

```
resources :orders, shallow: true do           
              resources :order_items
              get "status"
            end
```

reaching the resource directly seems to work as expected when trying to access a resource not owned by the user:

```
GET https://example.com/orders/10000

RESULTS IN: OrdersController action=show status=403 
```

however, if I try to access a custom action, unexpectedly I get back the result of that action on the unauthorized object:

```
GET https://example.com/orders/10000/status.json

RESULT: OrdersController action=status status=200
```

I can only get the expected behavior if I create my own load and authorize filter:

```
before_action :set_and_authorize_order

 def set_and_authorize_order
    @order = if params[:order_id]
               Order.find(params[:order_id])          
             end

    authorize! action_name.to_sym, @order
  end
```

I'm also able to access children of the order object with no issues:

```
GET https://example.com//orders/10000/order_items.json

RESULT: OrderItemsController action=index status=200
```

The request is actually handed off to the OrderItemsController not the OrdersController, I've checked and that controller in my code has an authorize_resource method at the top of the controller.

The ability model has the following rule:

```
 can [:show, :read, :create, :index], OrderItem do |orderitem|
        user.id == orderitem.order.user.id
      end
```

But this fails to block the request when the user is not an owner of the order in question...

### System configuration
**Rails version**: 6

**Ruby version**: 2.5

**CanCanCan version** 3.1.0

coorasse:
Hi! Thanks for pointing this out. Can you write a gist to reproduce the issue?

PR #767, opened 2022-02-04
Improve ability checks with nested resources (hash checks)

Juleffel:
When we use rails nested resources, cancancan does ability checks with `can? :action, { parent => Class }`. In this case, depending on how abilities are defined, we can have bugs and unexpected behaviours. I believe these are important issues that should be solved quickly.

I created [an example application](https://github.com/Juleffel/cancancan-test-jf) listing the problems arising depending on different ability configurations. You just need to clone it, seed it, and launch it. [The explanations are on the home page.](https://github.com/Juleffel/cancancan-test-jf/blob/master/app/views/home/index.html.erb)

[I also created a gist to show up the exact examples of can? that are failing.](https://gist.github.com/Juleffel/f44709b4fc116a473f8529576c784f9b)

I also believe issue #661 is related to this.


coorasse:
I am reading the examples and trying to understand what is going on

coorasse:
I am looking at the gist. In particular at this part:

```ruby
class AbilityCanID
  include CanCan::Ability

  def initialize(user)
    can :read, Book, user_id: [0, user.id]
  end
end
```

what I understand from here is "the user can read books with user_id `0` or `with my user_id` (they belong to me). So afterwards you do:

```ruby
user1 = User.create!
user2 = User.create!
Book.create([{user: user1}, {user: user2}])

ability = AbilityCanID.new(user1)
assert_equal true, ability.can?(:read, {user1 => Book})
assert_equal false, ability.can?(:read, {user2 => Book})
```

I am not entirely sure what you are checking with these assertions. Shouldn't you rather check:

```
user1 = User.create!
user2 = User.create!
book1, book2  = Book.create([{user: user1}, {user: user2}])

ability = AbilityCanID.new(user1)
assert_equal true, ability.can?(:read, book1)
assert_equal false, ability.can?(:read, book2)
```

?

coorasse:
I also looked into `/companies/3/project1s/custom_collection` from your Rails app.
What you say is:

> As it's not an owned company, we should not be able to see any projects.
> We should not be able to access this action.

The permissions are:

```ruby
can :read, Company
can :manage, Company, id: user.company_ids

can :read, Project1
can :manage, Project1, company_id: user.company_ids
```

1. what I read from these permissions is: "you can read companies, manage only yours". 
2. the controller checks against your "read" permissions for the company ([default](https://github.com/CanCanCommunity/cancancan/blob/develop/docs/nested_resources.md#nested-resources)) so it passes this check
3. the controller then checks the action `:custom_collection` (https://github.com/CanCanCommunity/cancancan/blob/develop/docs/changing_defaults.md#non-restful-controllers) but is not RESTful and therefore treats it as a collection and say: "give me all the project1 which I can `custom_collection` on", which is none due to `can :manage, Project1, company_id: user.company_ids`

If you want to have this action accessible only if you can manage a company, then you should check this manually in the action with `authorize! :manage, @company`.

coorasse:
Hi @Juleffel , thanks for the call today. I merged the first part of your changes on #772. Can you please rebase this PR so that it gets updated with the latest develop? Thanks

PR #780, opened 2022-03-29
Feature/cache subclasses

Liberatys:
Store the subclasses and only load them if they were not previously stored. 
This should make things run faster for applications that have a lot of different models with a big inheritance chain.

This will possibly fail if the application is changing the inheritance of classes during runtime.


coorasse:
@Liberatys is this PR tackling other open bugs? I remember something but I cannot find it anymore...

PR #799, opened 2022-07-13
Updating README(adding FAQ)

ImOwly:
This is just a suggestion but it adding a FAQ in the README can help people who use cancan to quickly resolve some problems that can occur the moment they use cancan instead of having to dig deeper into the doc everytime to find answers to this questions that frequently are asked, added format for example (just a suggestion)


coorasse:
Let's use the Discussion GitHub feature or open a new Markdown file. I'd like to keep the README as slim as possible. Thanks!

PR #806, opened 2022-11-17
use flat concatenation instead of nesting conditions

### LINKED ISSUE #805: Deep conditions nesting on sqlite => stack overflow
I am resubmitting this issue from the archived cancan project: [Deep conditions nesting on sqlite => stack overflow 
](https://github.com/ryanb/cancan/issues/414)

###

### Steps to reproduce
https://gist.github.com/NikolaiKircher/5f31bf2b0f9bf969597ce5e6e3f6551f

=> ActiveRecord::StatementInvalid: SQLite3::SQLException: parser stack overflow

### Actual behaviour
CanCan constructs sql like so

```
... where cond0 OR (
  cond1 OR (
    cond2 ...
  )
)
```
this useless deep nesting on sqlite3 results in a Parser Stack Overflow error.

### Expected behaviour
A better SQL would be:

`... where cond0 OR cond1 OR cond2 ...`

### System configuration
**Rails version**:
6.1.7
**Ruby version**:
3.0.4
**CanCanCan version**
3.4.0

NikolaiKircher:
closes #805 

**TODO:**
- [x] fix issue
- [ ] add tests?
- [ ] changelog

### DIFF
--- a/lib/cancan/model_adapters/active_record_adapter.rb
+++ b/lib/cancan/model_adapters/active_record_adapter.rb
@@ -157,7 +157,7 @@ module CanCan
         when false_sql
           behavior ? conditions : false_sql
         else
-          behavior ? "(#{conditions}) OR (#{sql})" : "not (#{conditions}) AND (#{sql})"
+          behavior ? "(#{conditions}) OR #{sql}" : "(not (#{conditions}) AND #{sql})"
         end
       end
 


coorasse:
Thanks for you PR. Could this lead to different results if the conditions are now "joined" without parenthesis?

NikolaiKircher:
> Thanks for you PR. Could this lead to different results if the conditions are now "joined" without parenthesis?

I think this change should be okay, but I can not guarantee it. (I just copied the solution from the archived cancan project: https://github.com/ryanb/cancan/issues/414 )

I just started the test suite and there are some failing examples, which would need to be adjusted. 
```
DB='sqlite' bundle exec appraisal activerecord_7.0.0 rake

537 examples, 12 failures, 3 pending

  expected: "not (\"articles\".\"secret\" = 1) AND ((\"articles\".\"published\" = 1) OR (\"articles\".\"id\" = 1))"
             got: "(not (\"articles\".\"secret\" = 1) AND (\"articles\".\"published\" = 1) OR \"articles\".\"id\" = 1)"

etc
```

I also found out, that the problem seems to be resolved with the latest version of sqlite3 (2.0.2) 
So maybe we don't even need this fix in the future

Issue #812, opened 2023-01-23
`can?` unnecessarily loads relationships in memory when it could leverage `accessible_by`


oboxodo:
### Steps to reproduce

You can find a full gist demonstrating the [problem and potential solution in here](https://gist.github.com/oboxodo/180703cff6182cdb63a9ba4091b59c26).

Given the following code:

```ruby
class Book < ActiveRecord::Base
  belongs_to :user
end

class User < ActiveRecord::Base; end

class Author < ActiveRecord::Base
  has_many :books
end

class Ability
  include CanCan::Ability

  def initialize(user)
    can :read, Author, books: { user: user }
  end
end
```

Asking if a user can read an author, shouldn't instantiate all the author's books in order to check the ability:

```ruby
author.books.loaded? # => false. This is good.
Ability.new(user).can?(:read, author)
author.books.loaded? # => true. This is BAD. There could be millions of records here.
```

Ideally, given all the defined abilities on `Author` are defined using the best practice of hash conditions, `can?` should be able to leverage `accessible_by` like this:

```ruby
Author.accessible_by(Ability.new(user), action).exists?(author.id)
author.books.loaded? # => false. Great.
```

### Expected behavior

The ability is checked without instantiating an unlimited number of records in memory.

### Actual behavior

Relationships involved in the hash conditions are [called](https://github.com/CanCanCommunity/cancancan/blob/7c99c593dd55127db16b0920167814a296c38e0d/lib/cancan/conditions_matcher.rb#L84) potentially instantiating an unlimited number of records in memory.

### System configuration

**Rails version**: 7.0.4
**Ruby version**: 3.0.2
**CanCanCan version**: 3.4.0

oboxodo:
I created a PR with a potential solution for this: https://github.com/CanCanCommunity/cancancan/pull/813

PR #813, opened 2023-01-23
Leverage accessible_by from can? when possible

oboxodo:
This is an attempt at solving https://github.com/CanCanCommunity/cancancan/issues/812.


oboxodo:
@coorasse do you think you can approve running the full tests on this PR?

PR #815, opened 2023-01-26
Renames method name in ConditionsMatcher

pietervisser:
Renames `ConditionsMatcher#matches_hash_conditions` to `matches_hash_conditions?` to better communicate that it returns a boolean, like the other `matches_xx?` methods.


coorasse (approved):
Thanks!

PR #818, opened 2023-02-16
Remove puts statement

phylor:
This removes a `puts` statement introduced in dcb429aef63910130811c7f212772b7fa918a926. I assume this wasn't left in intentionally?

It litters the output a bit when running tests or the app itself.


coorasse (approved):
Thanks!

PR #837, opened 2023-11-27
Extend member_action? by checking for instance_name id

0llirocks:
Currently only :id and :id_param are checked for member_action?. I think it makes sense to extend it by instance_name id.

Given the following url

/authors/1/books/1

the rails route might look like 

/authors/:id/books/:book_id or /authors/:author_id/books/:book_id

In these cases the member_action? will return false for either one or both controllers, since it won't find the :id param. This results in not loading the instance for actions like show/edit when using _authorize_resource_. The solution here is to explicity set the id_param option. But from my point of view, this id naming scheme is quite common and should be considered when checking the id param.

This PR adds an id check for instance_name + _id, which would fullfil most of these cases.

### DIFF
--- a/lib/cancan/controller_resource.rb
+++ b/lib/cancan/controller_resource.rb
@@ -80,7 +80,7 @@ module CanCan
 
     def member_action?
       new_actions.include?(@params[:action].to_sym) || @options[:singleton] ||
-        ((@params[:id] || @params[@options[:id_param]]) &&
+        ((@params[:id] || @params["#{instance_name}_id"] || @params[@options[:id_param]]) &&
           !collection_actions.include?(@params[:action].to_sym))
     end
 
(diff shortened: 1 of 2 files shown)


itadventurer:
I was also on the search for this. For me, this was not enough, I needed to also patch the `id_param` method of the `CanCan::ControllerResourceLoader`.
My quick and dirty monkeypatch looks the following way:

```ruby
module PatchCanCanControllerResource
  def member_action?
    new_actions.include?(@params[:action].to_sym) || @options[:singleton] ||
      ((@params[:id] || @params["#{instance_name}_id"] || @params[@options[:id_param]]) &&
        !collection_actions.include?(@params[:action].to_sym))
  end
end

CanCan::ControllerResource.prepend PatchCanCanControllerResource

module PatchCanCanControllerResourceLoader
  def id_param
    return @params[id_param_key].to_s if @params[id_param_key].present?
    return @params["#{instance_name}_id"].to_s if @params["#{instance_name}_id"].present?
  end
end

CanCan::ControllerResourceLoader.prepend PatchCanCanControllerResourceLoader
```

Issue #838, opened 2023-11-30
Support Rails's `attribute` in Cancancan's `permitted_attributes`


kalsan:
TLDR: Cancancan does not "see" non-database backed attributes, so they are missing in `permitted_attributes`.

### Steps to reproduce

Rails 7.1.2 application with a basic user model with the database fields `email` and `password_digest`, with the important parts from the model being:

```ruby
class User < ApplicationRecord

has_secure_password
```

With the god-ability:

```
class Ability
  include CanCan::Ability

  attr_reader :user

  def initialize(user)
    return if user.blank?

    can :manage, :all
  end
end
```

Now run: `Ability.new(User.first).permitted_attributes(:edit, User.first)`

### Expected behavior

```ruby
[:email, :password_digest, :password, :password_confirmation]
```

### Actual behavior

```ruby
[:email, :password_digest]
```

-> Cancancan misses out on the password attributes. The same problem can be reproduced by using Rail's `attribute` method, i.e.

```ruby
class MyModel < ApplicationRecord
  attribute :foo # Cancancan will miss this
end
```

### System configuration
**Rails version**: 7.1.2

**Ruby version**: 3.2.2

**CanCanCan version** 3.4.0

I guess this also interferes with other Gems such as `active_type` or any custom accessor logic using virtual fields using `attribute`.

I hope the examples above are useful to you. Let me know if I can help you enlighten this further.

Best,
Kalsan

PR #839, opened 2023-12-01
In permitted_attributes, switch from database column detection to Rails attributes detection. Addresses #838

kalsan:
This pull request addresses #838 and makes Cancancan "see" attributes that are not backed by the database. It enables code like the following:

```ruby
class Foo < ApplicationRecord
  attribute :bar
end
```

With this pull request, cancancan's `permitted_attributes` will "see" and include `:bar` even when there is no corresponding column in the database.

Reasonning why this is a good idea: When implementing a form and checking for per-attribute authorization, there are often more fields in the form than are actually stored. An example would be a checkbox "I agree to the terms of use" which is checked by the controller and then discarded if checked, validation error otherwise. If cancancan ignores this attribute (since it's not DB backed), a form whitelisting attributes by using cancancan's `permitted_attributes` would miss on this field.

This becomes particularly important when using cancancan together with Gems like `active_type`.

### DIFF
--- a/lib/cancan/ability/strong_parameter_support.rb
+++ b/lib/cancan/ability/strong_parameter_support.rb
@@ -31,7 +31,7 @@ module CanCan
         klass = subject_class?(subject) ? subject : subject.class
         # empty attributes is an 'all'
         if rule.attributes.empty? && klass < ActiveRecord::Base
-          klass.column_names.map(&:to_sym) - Array(klass.primary_key)
+          klass.attribute_names.map(&:to_sym) - Array(klass.primary_key)
         else
           rule.attributes
         end
--- a/lib/cancan/ability/strong_parameter_support.rb
+++ b/lib/cancan/ability/strong_parameter_support.rb
@@ -31,7 +31,7 @@ module CanCan
         klass = subject_class?(subject) ? subject : subject.class
         # empty attributes is an 'all'
         if rule.attributes.empty? && klass < ActiveRecord::Base
-          klass.attribute_names.map(&:to_sym) - Array(klass.primary_key)
+          klass.column_names.map(&:to_sym) - Array(klass.primary_key)
         else
           rule.attributes
         end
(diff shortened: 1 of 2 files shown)


kalsan:
Hi @coorasse, it was a pleasure meeting you IRL :-)

Thank you for your suggestions regarding testing. You've asked me to implement a failing "before" test, as well as a passing "after" test. However, it's a bit trickier than that because of the following situation:

- The current test verifying that column names are read correctly by CanCanCan creates a method `column_names` on an `ActiveRecord::Base` and checks that they are read back correctly.
- However, this method does not affect `attribute_names` which is inconsistent with Rails' actual behavior: when loading the columns, Rails automatically adds each column under `attribute_names`, see also [this StackOverflow question regarding the topic](https://stackoverflow.com/questions/44967718/what-is-the-difference-between-an-attribute-and-a-column).
- Therefore, while this PR works in real life, the existing test would fail.

Therefore, I modified the PR to consist of three commits:

- [1edfe71](https://github.com/CanCanCommunity/cancancan/pull/839/commits/1edfe7158ea353fb15a3df4a3f974f419269d7ea) modifies the test to actually perform testing on the DB instead of working only in-memory, enabling the needed Rails behavior of propagating column names to attributes.
- [4936716](https://github.com/CanCanCommunity/cancancan/pull/839/commits/49367167516af1d3e7204f5ff40ee1e690a1fc90) modifies the test such that it will fail with the current implementation, demonstrating that attributes are not correctly handled in 3.5.0.
- [f68d752](https://github.com/CanCanCommunity/cancancan/pull/839/commits/f68d752f735d3af7a201b41fd34a5f96d1678ac6) adds the fix that makes the test pass. This is the commit this PR is for.

I hope that these suggested code changs are to your liking and that the PR can be merged to improve CanCanCan as discussed :-)

Have a nice prolonged week-end!
Best,
Kalsan

PR #847, opened 2024-04-30
Fix: rule_spec should honor DB setting

tardate:
I noticed when running db-specific tests e.g.
```
DB='postgres' bundle exec appraisal activerecord_6.1.0 rake
```
that `spec/cancan/rule_spec.rb` could fail if sqlite not available, because it was ignoring the DB context and always/only running with sqlite. 

This is just a quick fix to ensure that `spec/cancan/rule_spec.rb` runs like all other tests with the specific DB selected.


coorasse (approved):
Thank you!

PR #849, opened 2024-05-23
Update the tests matrix

coorasse:
Let's refresh the matrix of tests, removing ruby 2.6 and rails < 5.2.
We also add rails 7.0 and 7.1 and ruby 3.2 and 3.3.
We stop testing on jruby and truffleruby because tests are failing right now


.github/workflows/test.yml
```diff
@@ -14,41 +14,38 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: ['2.6', '2.7', '3.0', 'jruby', 'truffleruby']
-        gemfile: ['gemfiles/activerecord_5.0.2.gemfile', 'gemfiles/activerecord_5.1.0.gemfile', 'gemfiles/activerecord_5.2.2.gemfile', 'gemfiles/activerecord_6.0.0.gemfile', 'gemfiles/activerecord_6.1.0.gemfile', 'gemfiles/activerecord_main.gemfile']
+        ruby: ['2.7', '3.0', '3.1', '3.2', '3.3']
```
andrykonchin, line 17: > We stop testing on jruby and truffleruby because tests are failing right now

It seems specs should pass on TruffleRuby now (or actually fail in the same way they are failing on CRuby on the `develop` branch).

I've enabled truffleruby-head in a fork (https://github.com/andrykonchin/cancancan/pull/1) and there is only one failing test case that fails on CRuby as well now:
- https://github.com/andrykonchin/cancancan/actions/runs/9464710110/job/26072611802?pr=1
- https://github.com/andrykonchin/cancancan/actions/runs/9464710110/job/26072611600?pr=1
- https://github.com/andrykonchin/cancancan/actions/runs/9464710110/job/26072611391?pr=1
- https://github.com/andrykonchin/cancancan/actions/runs/9464710110/job/26072611180?pr=1
- https://github.com/andrykonchin/cancancan/actions/runs/9464710110/job/26072610978?pr=1

TruffleRuby failures on CI (before it was disabled) were expected and related to cancancan's depending on the old `rake` gem version (10.5.0) that doesn't support Ruby 3.2 because of a breaking change in Ruby 3.2 - removing `Object#=~` method (TruffleRuby's latest releases target Ruby 3.2):

```
Run DB=sqlite bundle exec rake
rake aborted!
NoMethodError: undefined method `=~' for #<Proc:0x2c08>
<internal:core> core/enumerable.rb:501:in `block in sort_by'
<internal:core> core/enumerable.rb:500:in `map'
<internal:core> core/enumerable.rb:500:in `sort_by'
<internal:core> core/kernel.rb:378:in `load'
<internal:core> core/kernel.rb:378:in `load'
<internal:core> core/kernel.rb:378:in `load'
/home/runner/.rubies/truffleruby-24.0.1/bin/bundle:44:in `<main>'
(See full trace by running task with --trace)
```

The Rake issue was fixed in some recent release (https://github.com/ruby/rake/commit/81763da40dcfc497e25e00a7e957efd84a053923). Now specs are run without involving Rake (with `bundle exec rspec`) so the Rake issue doesn't affect it. Specs weren't failing on CRuby on CI in the same time when were failing on TruffleRuby because they weren't run on Ruby 3.2 and 3.3 that have the breaking change.

cc @eregon

andrykonchin: Created a PR #853

PR #856, opened 2024-07-15
"Cannot" with association hash conditions : accessible_by generates incorrect SQL (Postgres / ActiveRecord)

0llirocks:
I hope it's enough to link to the relevant issue #214 

The commit addresses the issue at its root by checking for _IS NULL_. PostgreSQL can't compare null values with any other value.


0llirocks:
@coorasse Hi, you seem to be one of the maintainer. Any updates on this PR? It is still an issue in our applications that we can't really fix otherwise.

Best regards

PR #877, opened 2025-03-02
Check rule's subject's inheritance if subject is a Class or Module

kevinluo201:
This change fixes the following problem of STI classes

```ruby
# given "class Child < Parent"
can :read, Parent
cannot :read, Child

can? :read, Parent # => false, but should be true
```
The reason is that `relevant_rules` will fetch both ancestors and subclasses rules when ActiveRecord is used, where non-sti classes don't have this problem.

The fix is simple: just check if the rule's subject is a ancestor or itself

### DIFF
--- a/lib/cancan/conditions_matcher.rb
+++ b/lib/cancan/conditions_matcher.rb
@@ -7,6 +7,7 @@ module CanCan
       return call_block_with_all(action, subject, extra_args) if @match_all
       return matches_block_conditions(subject, attribute, *extra_args) if @block
       return matches_non_block_conditions(subject) unless conditions_empty?
+      return subject_descendant?(subject) if subject_class?(subject)
 
       true
     end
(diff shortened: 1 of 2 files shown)


kevinluo201:
Hi @coorasse, this is my first time contributing to this project. Would you mind helping me review this PR? Does it need to pass any CI? Thank you 🙏

kevinluo201:
@brynary @macav @Senjai @amitsuryavanshi Sorry to bother you 🥹 I just want to ask if anyone can review this PR if they're available. 🙏

dalezak:
@coorasse any ETA when this PR will be merged? I'm running into this issue on a project, trying to decided whether to monkey patch or wait for this to be merged. /cc @kevinluo201

23tux:
@kevinluo201 any updates on this?

PR #878, opened 2025-03-07
Support `through_association` with `singleton`.

### LINKED ISSUE #604: `:through_association` and `:singleton`
https://github.com/CanCanCommunity/cancancan/blob/6d0e62bf0c475964a0533608e7e7221be45de28c/lib/cancan/controller_resource_finder.rb#L8

Just ran into an issue here. If I pass `:through_association` and `:singleton` things don't work.
It should check and send `@options[:through_association]` if set, I guess?

cgunther:
Given a load declaration like:
```ruby
load_resource :order
load_resource through: :address, through_association: :custom_address, singleton: true
```

The `:through_assocation` option wasn't being respected in the `:singleton` branch, thus falling back to calling `find`, but odds are you wouldn't have an ID given it was a singleton, thus leading to an error.

Now it uses the `:through_association` before falling back to the `name`.

Fixes #604.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+* [#878](https://github.com/CanCanCommunity/cancancan/pull/878): Support `through_association` with `singleton`. ([@cgunther][])
+
 ## 3.6.0
 
 * [#849](https://github.com/CanCanCommunity/cancancan/pull/849): Update tests matrix. ([@coorasse][])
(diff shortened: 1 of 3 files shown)


cgunther:
CI seems to be failing against ActiveRecord main due to bumping to sqlite3 v2 in Rails 8.0 but the gemfile is using v1.7:
https://github.com/rails/rails/commit/2976d3767e6572ee1838b50b5f401983918f99c7

It's also failing against ActiveRecord v6.0, v6.1, and v7.0 due to a change in concurrent-ruby where it stopped requiring the logger gem, but ActiveSupport was relying on it to require logger. This was fixed in v7.1:
https://github.com/rails/rails/pull/49372
and a PR was merged to branch 7-0-stable and 6-1-stable, but there hasn't been a release yet:
https://github.com/rails/rails/pull/54264
https://github.com/rails/rails/pull/54621

Issue #889, opened 2026-03-09
teach condition_match? to handle ActiveRecord::Relation values


apneadiving:
I ended up patching cancancan for the following usecase: teach condition_match? to handle ActiveRecord::Relation values.

Would it make sense to add it to the gem itself?

CONTEXT
-------
CanCanCan ability conditions accept both arrays and AR relations as condition values:
```
  can :read, UserContract, team_id: [1, 2, 3]          # Array     → IN list in SQL
  can :read, UserContract, team_id: Team.select(:id)   # Relation  → subquery in SQL
```

We use AR::Relation to avoid IN lists with hundreds of bind parameters, which degrade query plan quality and prevent plan caching.

THE BUG
-------
CanCanCan's condition_match? is used for two things:
  1. accessible_by(ability) — builds SQL scopes. AR::Relation works perfectly here.
  2. can?(permission, instance) — checks a single object in Ruby. Broken for relations.

The original code in conditions_matcher.rb:
```
  when Enumerable then value.include?(attribute)
```

AR::Relation includes Enumerable, so it falls into this branch. But Enumerable#include?
iterates over the relation records and compares each with ==. Since the relation yields
AR objects (e.g., Team instances) and attribute is a primitive (e.g., integer team_id),
Team.new(id: 5) == 5 is always false — every instance-level can? check returns false.

THE FIX
-------
Prepend a module that intercepts condition_match? when value is an AR::Relation,
using exists? with the model's primary key instead of Enumerable#include?.
All our subquery methods select(:id) from the canonical model, so primary_key is
always 'id' and value.where(pk => attribute).exists? is always correct.

```
module CanCanCanArRelationConditionMatch
  private

  def condition_match?(attribute, value)
    if value.is_a?(ActiveRecord::Relation)
      # See comment above. Must be checked before Enumerable since AR::Relation
      # includes Enumerable but Enumerable#include? compares objects, not IDs.
      value.where(value.primary_key => attribute).exists?
    else
      super
    end
  end
end
CanCan::ConditionsMatcher.prepend(CanCanCanArRelationConditionMatch)
```

PR #890, opened 2026-04-06
Fix can? for ActiveRecord relation-based conditions

olistik:
This PR fixes an inconsistency between `accessible_by` and `can?` when an ability condition uses an `ActiveRecord::Relation` as the condition value.

Example:

```ruby
can :read, Article, id: Article.where(user_id: user.id).select(:id)
```

Before this change:

* `Article.accessible_by(ability)` correctly returns the expected records
* `ability.can?(:read, article)` incorrectly returns `false`

After this change:

* `accessible_by` continues to behave correctly
* `can?` now returns the expected result for matching records

---

### Problem

`accessible_by` works because ActiveRecord conditions are translated into SQL.

However, `can?` performs an in-memory condition match, and when the condition value is an `ActiveRecord::Relation`, it was comparing the record attribute against relation objects rather than against the selected scalar values.

This caused behavior like:

```ruby
user = User.create!(name: 'Arthur Dent')
article = Article.create!(name: 'How to fly', user:)

ability = Ability.new(user)
ability.can :read, Article, id: Article.where(user_id: user.id).select(:id)

Article.accessible_by(ability) # => includes article
ability.can?(:read, article)   # => false (before)
ability.can?(:read, article)   # => true  (after)
```

---

### Change

This PR updates `CanCan::ConditionsMatcher#condition_match?` so that when a condition value is an `ActiveRecord::Relation`, it is normalized into the selected record values before performing the in-memory match.

This makes `can?` consistent with `accessible_by` for relation-backed conditions such as:

```ruby
id: Article.where(user_id: user.id).select(:id)
```

---

### Test coverage

Added a spec in:

```ruby
spec/cancan/model_adapters/active_record_adapter_spec.rb
```

The example verifies that:

* `accessible_by` returns the expected matching record
* `can?` returns `true` for the matching record
* `can?` returns `false` for a non-matching record

---

### Why this matters

Using subqueries / relation-backed conditions is already a valid and useful pattern in ability definitions, especially when trying to avoid eager-loading IDs into Ruby arrays.

This change ensures that:

* SQL-backed authorization (`accessible_by`)
* in-memory authorization (`can?`)

follow the same logic and produce consistent results.

---

### Notes for reviewers

This change is intentionally narrow and only affects in-memory condition matching when a condition value is an `ActiveRecord::Relation`.

It does not change SQL generation or `accessible_by` behavior, which was already correct.

### DIFF
--- a/lib/cancan/conditions_matcher.rb
+++ b/lib/cancan/conditions_matcher.rb
@@ -88,6 +88,9 @@ module CanCan
     end
 
     def condition_match?(attribute, value)
+      if defined?(ActiveRecord) && value.is_a?(ActiveRecord::Relation)
+        value = value.map { |v| v.attributes.compact_blank.values[0] }
+      end
       case value
       when Hash
         hash_condition_match?(attribute, value)
--- a/spec/cancan/model_adapters/active_record_adapter_spec.rb
+++ b/spec/cancan/model_adapters/active_record_adapter_spec.rb
@@ -276,6 +276,32 @@ RSpec.describe CanCan::ModelAdapters::ActiveRecordAdapter do
         expect(Article.accessible_by(@ability)).to match_array([article1])
       end
 
+      it 'matches records for can? when a condition value is an ActiveRecord relation selecting ids' do
+        user = User.create!(name: 'Arthur Dent')
+        article = Article.create!(name: 'How to fly', user: user)
+        other_article = Article.create!(name: 'Mostly Harmless')
+        ability = Ability.new(user)
+
+        ability.can :read, Article, id: Article.where(user_id: user.id).select(:id)
+
+        expect(Article.accessible_by(ability)).to match_array([article])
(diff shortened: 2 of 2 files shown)


olistik:
Fixes #889

coorasse (commented):
I am very surprised that this was not covered already. Thank you. I'd love to see also a test covering the same but where we don't use `id` but, for example, `name` as a condition. This would make it clear that is not something specific for the primary key, but for any other column as well. Basically add a second test where we do the same check but with the ability

`ability.can :read, Article, name: Article.where(user_id: user.id).select(:name)`

olistik:
@coorasse I can see the reason to add that second test example.
I added it [here](https://github.com/CanCanCommunity/cancancan/pull/890/changes#diff-17131cd8d61b3445e02144eac7a7ed2c3215bf4b16f41875d665cbb4693e624aR292-R303) and then rebased. 🎩

PR #892, opened 2026-05-25
fix: prevent accessible_by from permanently overwriting shared rule conditions

### LINKED ISSUE #876: accessible_by permanently modifies a rule's conditions via normalize_conditions, causing can? to fail
The ConditionNormalizer in accessible_by permanently modifies a rule's conditions when you call accessible_by, meaning that a call to the can? method that worked before calling accessible_by may not work afterwards.

For example, say I have the models Client, Project, and Activity, where an activity belongs to a project and a project belongs to a client. In Ability, I can create a rule to say that a user can read any project or activity belonging to client with id 10:

`can :read, [ Project, Activity ], client: { id: 10 }`

This creates a rule with conditions: { client: { id: 10 } }.

Then if I run this code in the controller:
```
    @project = Project.first
    can? :read, @project
    @activities = Activity.accessible_by current_ability
    can? :read, @project

```
then the first call to can? runs without error. The call to Activity.accessible_by current_ability permanently changes the rule's conditions to { project: { client: { id: 10 } } } via a call to normalize_conditions, and the second identical call to can? then raises a NoMethodError saying "undefined method `project' for an instance of Project" because the new conditions make cancancan call @project.project.client instead of @project.client.

Is the rule illegal, ie should you only ever create a rule with an array of models that all have through reflections to the condition OR all have belongs_to reflections to the condition? Or should accessible_by not permanently change the conditions for the rule?

I'm using cancancan 3.6.1 and Rails 8.0.1.

SAY-5:
## What does this fix?

Calling `accessible_by` for one model class permanently overwrites `rule.conditions` on the **shared** `Rule` objects stored in the ability's `@rules`. A subsequent `can?` check for a different model that shares the same rule then uses the already-normalized conditions, which may reference associations that do not exist on that model, causing `NoMethodError` or wrong results.

### Reproduction (from #876)

```ruby
# Ability definition - rule applies to both Project and Activity
can :read, [Project, Activity], client: { id: current_client.id }

# Project belongs_to :client (direct)
# Activity has_many-through path to client via :project

# Step 1 - works fine before accessible_by is called
can?(:read, @project)  # => correct result

# Step 2 - triggers ConditionsNormalizer for Activity
# Rewrites rule.conditions from { client: { id: x } }
# to { project: { client: { id: x } } } on the SHARED Rule object
Activity.accessible_by(current_ability)

# Step 3 - now broken
can?(:read, @project)  # => NoMethodError: undefined method `project' for Project
```

## Fix

`ActiveRecordAdapter#initialize` now maps `@compressed_rules` through `Rule#dup` before passing to `StiNormalizer` and `ConditionsNormalizer`. The dup gives each adapter its own copy of the rule shell, so `conditions=` assignments from the normalizers are local and never touch the originals in `@rules`.

```ruby
@compressed_rules = base_rules.map(&:dup)
```

## Tests

Added a regression spec to `conditions_normalizer_spec.rb`:

- Confirms `rule.conditions` is **unchanged** after `accessible_by` is called (object identity preserved)
- Fails without the fix (rule.conditions becomes the normalized form)
- Passes with the fix

All 305 examples pass (0 failures, 2 pending) across both defined and random order.

Closes #876

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+* [#876](https://github.com/CanCanCommunity/cancancan/issues/876): Fix `accessible_by` permanently overwriting shared rule conditions, causing `can?` to fail for models sharing the same rule. ([@say-apm35][])
+
 ## 3.6.0
 
 * [#849](https://github.com/CanCanCommunity/cancancan/pull/849): Update tests matrix. ([@coorasse][])
--- a/lib/cancan/model_adapters/active_record_adapter.rb
+++ b/lib/cancan/model_adapters/active_record_adapter.rb
@@ -18,11 +18,18 @@ module CanCan
 
       def initialize(model_class, rules)
         super
-        @compressed_rules = if CanCan.rules_compressor_enabled
-                              RulesCompressor.new(@rules.reverse).rules_collapsed.reverse
-                            else
-                              @rules
-                            end
+        base_rules = if CanCan.rules_compressor_enabled
+                       RulesCompressor.new(@rules.reverse).rules_collapsed.reverse
+                     else
+                       @rules
+                     end
+        # Dup each rule so that normalization for this model class does not permanently
+        # alter the shared rule objects. ConditionsNormalizer replaces rule.conditions with
+        # a newly computed hash, which overwrites the shared rule's conditions and corrupts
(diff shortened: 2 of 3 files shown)


SAY-5:
Anything blocking here? accessible_by mutates the rule's conditions hash in place while merging, so the rule stays modified for every later check in the same process. Fix copies before merging.

